PR #23, opened 2023-06-27
Initial/rest controllers generators

alexandervlasenko:


ILoveMyTortoise (approved):

src/Nox.Generator/Presentation/Api/ApiGenerator.Rest/ApiControllerGenerator.cs
rochar, line 1: @alexandervlasenko can we add a test to see if the class is generated, similar to what we have?

rochar: Can you provide a screen shoot of the generated class, so its easier to review, thks

rochar (changes requested):
looks good, just a comment on tests and be able to see the generated class

src/Nox.Generator/Presentation/Api/ApiGenerator.Rest/ApiControllerGenerator.cs
alexandervlasenko: Hi @rochar , I've added a new test and also a sample file into examples folder (src/SampleWebApp/Examples/CountriesApiController.cs) so no need in a screenshot.

Best regards

rochar (approved):

jan-schutte (approved):

PR #29, opened 2023-06-30
enable database configuration per database provider (Text and Number Only)

rochar:
implement Nox.Type Text
Add this code to Nox.Generators is ok since we do not have any third party dependency, we are only generating strings, this accelerates development and we can move this to its on component if necessary

we may implement a DefaultConfig, would work generic in each db

we need to implement explicitly for each Nox.Type, bust this is centralized and encapsulated

### DIFF
--- a/src/Nox.Generator/Infrastructure/Persistence/DatabaseConfiguration/ConfigurationDefaults.cs
+++ b/src/Nox.Generator/Infrastructure/Persistence/DatabaseConfiguration/ConfigurationDefaults.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using Nox.Types;
+
+namespace Nox.Generator.Infrastructure.Persistence.DatabaseConfiguration;
+
+internal static class ConfigurationDefaults
+{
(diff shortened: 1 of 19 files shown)


ILoveMyTortoise (approved):

src/Nox.Generator/Infrastructure/Persistence/DatabaseConfiguration/ConfigurationDefaults.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Nox.Generator.Infrastructure.Persistence.DatabaseConfiguration;
+
+internal static class ConfigurationDefaults
+{
+    public static DatabaseAttributeConfig GetDefaultOptions(TextTypeOptions options)
+    {
+        return new DatabaseAttributeConfig()
+        {
+            IsSingleProperty = true,
+            HasConversionTypeFullName = "Nox.Types.EntityFramework.Types.TextConverter",
```
ILoveMyTortoise, line 13: why not just ConversionTypeFullName

jan-schutte (approved):

alexandervlasenko (approved):

PR #158, opened 2023-07-05
CurrencyCode Implementation

denisvieira-dev:


src/Nox.Types/Types/CurrencyCode3/CurrencyCode3.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Linq;
+
+namespace Nox.Types;
+
+/// <summary>
+/// Class for three-letters currency code (ISO 4217).
+/// </summary>
+public sealed class CurrencyCode3 : ValueObject<string, CurrencyCode3>
+{
+    private readonly string[] CurrencyCodes = Enum.GetNames(typeof(CurrencyCode));
```
dejobratic, line 11: Can we put this as a private readonly static variable? Also, can we rename it using the following convention https://regusit.atlassian.net/wiki/spaces/NOX/pages/3153461266/Coding+Conventions+Styles+and+Guidelines#Private-static-fields?

dejobratic (changes requested):
Looks good to me. Just a small comment.

tests/Nox.Types.Tests/Types/CurrencyCode3/CurrencyCode3Tests.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        var currencyCode3_1 = CurrencyCode3.From("USD");
+
+        var currencyCode3_2 = CurrencyCode3.From("USD");
+
+        currencyCode3_1.Should().Be(currencyCode3_2);
+    }
+
+    [Fact]
+    public void CurrencyCode3_CaseInsensitiveEquality_Tests()
```
rochar, line 35: this test is failing in the CI, but should be valid / its correctly implemented.

I think we should do a ToUpper in From  method because we assume that Currency is always in upper case, do y agree?

rochar (changes requested):
Looks good, small comment

tests/Nox.Types.Tests/Types/CurrencyCode3/CurrencyCode3Tests.cs
denisvieira-dev: I totally agree. Updating them

src/Nox.Types/Types/CurrencyCode3/CurrencyCode3.cs
denisvieira-dev: Done

dejobratic (changes requested):
Looks good. Just added a comment related to type name.

src/Nox.Types/Types/CurrencyCode3/CurrencyCode3.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Linq;
+
+namespace Nox.Types;
+
+/// <summary>
+/// Class for three-letters currency code (ISO 4217).
+/// </summary>
+public sealed class CurrencyCode3 : ValueObject<string, CurrencyCode3>
```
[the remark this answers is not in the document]
denisvieira-dev: Hi Dejan, thanks for your suggestion. I agree, but we have an Enum called CurrencyCode. Should we rename this classe for CurrencyCode2 or give a prefix for that enum?

rochar (changes requested):

src/Nox.Types/Types/CurrencyCode3/CurrencyCode3.cs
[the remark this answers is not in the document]
dejobratic: Thanks for the info @rochar. I'll resolve this comment.

rochar (approved):

dejobratic (approved):

PR #166, opened 2023-07-06
MP-7343 [BS-8103] NOX "DateTime" - Define and implement custom type

becicajla:


src/Nox.Types/Types/DateTime/DateTime.cs
jan-schutte, line 1: This should be a sealed class

src/Nox.Types/Types/DateTime/DateTime.cs
```diff
@@ -0,0 +1,198 @@
...
+        get { return Value.Second; }
+        private set { Value = new System.DateTime(Value.Year, Value.Month, Value.Day, Value.Hour, Value.Minute, value); }
+    }
+
+    public DateTime() { Value = System.DateTime.MinValue; }
+
+    public static DateTime From(int year, int month, int day, int hour = 0, int minute = 0, int second = 0) => From(new YearTypeOptions(), year, month, day, hour, minute, second);
+
+
+    public static DateTime From(YearTypeOptions options, int year, int month, int day, int hour = 0, int minute = 0, int second = 0)
```
rochar, line 44: The YearTypeOptions should only be used by Year Type is not designed to be used in DateTime

We need to define a good DateTimeOptions (Min. Max) and build a proper Validation without using the other types, should use .Net Datetime only

Lets keep this very simple

src/Nox.Types/Types/DateTime/DateTime.cs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        Value = Value.Add(ts);
+        return this;
+    }
+
+    public override bool Equals(object obj)
+    {
+        return obj is DateTime time &&
+               base.Equals(obj) &&
+               Value == time.Value &&
```
rochar, line 184: The Value underlinetype is Datetime, so doing a equal on the value should be enough:

` return obj is DateTime time &&
               base.Equals(obj) && Value.Equals(time.Value) `

rochar (changes requested):
Some comments to validate

src/Nox.Types/Types/DateTime/DateTime.cs
```diff
@@ -0,0 +1,198 @@
...
+        private set { Value = new System.DateTime(Value.Year, Value.Month, Value.Day, Value.Hour, value, Value.Second); }
+    }
+    public int Second
+    {
+        get { return Value.Second; }
+        private set { Value = new System.DateTime(Value.Year, Value.Month, Value.Day, Value.Hour, Value.Minute, value); }
+    }
+
+    public DateTime() { Value = System.DateTime.MinValue; }
+
```
rochar, line 40: Consider add a From overload that receives a System.DateTime

src/Nox.Types/Types/DateTime/DateTime.cs
```diff
@@ -0,0 +1,275 @@
...
+    /// </summary>
+    /// <param name="year"></param>
+    /// <param name="month"></param>
+    /// <param name="day"></param>
+    /// <param name="hour"></param>
+    /// <param name="minute"></param>
+    /// <param name="second"></param>
+    /// <returns>New <see cref="DateTime"/> object with default <see cref="DateTimeTypeOptions"/> .</returns> 
+    /// <exception cref="TypeValidationException"></exception>
+    public static DateTime From(int year, int month, int day, int hour = 0, int minute = 0, int second = 0) => From(new DateTimeTypeOptions(), year, month, day, hour, minute, second);
```
gjorgji-grgovski-authoritypartners, line 54: There is no need to have these overloads, as you can use the
`From(System.DateTime dateTime)`
and just instantiate System.DateTime based on year, month, and day.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types/Types/DateTime/DateTime.cs
danield-ap: Overload with the dates as integers removed.

src/Nox.Types/Types/DateTime/DateTime.cs
```diff
@@ -0,0 +1,275 @@
...
+    /// <param name="options"></param>
+    /// <param name="year"></param>
+    /// <param name="month"></param>
+    /// <param name="day"></param>
+    /// <param name="hour"></param>
+    /// <param name="minute"></param>
+    /// <param name="second"></param>
+    /// <returns>New <see cref="DateTime"/> object with sent <see cref="DateTimeTypeOptions"/>.</returns> 
+    /// <exception cref="TypeValidationException"></exception>
+    public static DateTime From(DateTimeTypeOptions options, int year, int month, int day, int hour = 0, int minute = 0, int second = 0)
```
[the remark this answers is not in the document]
danield-ap: Overload with the dates as integers removed.

src/Nox.Types/Types/DateTime/DateTime.cs
```diff
@@ -0,0 +1,275 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace Nox.Types;
+public sealed class DateTime : ValueObject<System.DateTime, DateTime>
+{
+    private DateTimeTypeOptions _dateTimeTypeOptions = new();
+
+    public int Year
```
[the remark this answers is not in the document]
danield-ap: Properties removed.

src/Nox.Types/Types/DateTime/TypeOptions/DateTimeTypeOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿
+namespace Nox.Types;
+public class DateTimeTypeOptions
```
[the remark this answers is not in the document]
danield-ap: Schema test was run and file changes committed.

gjorgji-grgovski-authoritypartners (approved):

src/Nox.Types/Types/DateTime/DateTime.cs
```diff
@@ -0,0 +1,176 @@
...
+        return Value.ToString(format, formatProvider);
+    }
+
+    /// <summary>
+    /// Adds sent Timespan on <see cref="DateTime.Value"/>
+    /// </summary>
+    /// <param name="ts"></param>
+    public void Add(TimeSpan ts)
+    {
+        Value = Value.Add(ts);
```
rochar, line 173: Nox.Types are immutable, needs to create a new Instance like From, can be marked as static

src/Nox.Types/Types/DateTime/TypeOptions/DateTimeTypeOptions.cs
rochar, line 1: Add a property to NoxSimpleTypeDefinition,see other type options and regenerate the Nox.Solution schemas

src/Nox.Types/Types/DateTime/TypeOptions/DateTimeTypeOptions.cs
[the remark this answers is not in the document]
rochar: I do not see the new schemas?

rochar (changes requested):
just one important comment regarding Nox.Types, and another regarding schemas

src/Nox.Types/Types/DateTime/DateTime.cs
becicajla: Add removed. It was added in case DateTimeRange uses Nox.Types.DateTime, but now it is redundant since we have operator +

rochar (approved):

PR #176, opened 2023-07-06
MP-7399: Json type definition and implementation.

originalethanhunt:


src/Nox.Types/Types/Json/Json.cs
```diff
@@ -0,0 +1,114 @@
...
+
+/// <summary>
+/// Represents a Nox <see cref="Json"/> type and value object.
+/// </summary>
+public sealed class Json : ValueObject<string, Json>
+{
+    /// <summary>
+    /// Gets or sets the options for processing json.
+    /// </summary>
+    public JsonTypeOptions Options { get; set; } = new JsonTypeOptions();
```
dejobratic, line 14: Should the options be a private field?

src/Nox.Types/Types/Json/Json.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            JsonDocument.Parse(Value);
+        }
+        catch (JsonException ex)
+        {
+            result.Errors.Add(new ValidationFailure(nameof(Value), $"Could not create a Nox Json type with value {Value} due to JsonException ({ex.Message})."));
+        }
+        catch (Exception ex)
+        {
+            result.Errors.Add(new ValidationFailure(nameof(Value), $"Could not create a Nox Json type with value {Value} due to Exception ({ex.Message}).\""));
```
dejobratic, line 85: We have a trailing double quote here.

dejobratic (changes requested):
Looks good to me. Just added a few small comments.

src/Nox.Types/Types/Json/Json.cs
```diff
@@ -0,0 +1,114 @@
...
+                base.Value = _returnValue = value;
+        } 
+    }
+
+    /// <inheritdoc/>
+    internal override ValidationResult Validate()
+    {
+        var result = base.Validate();
+
+        try
```
dejobratic, line 75: For other types we didn't expose the inner exception message, so we should check this with client or go with the standard message format, like:

Could not create a Nox Json type because {Value} does not have a valid Json format.

andresharpe (approved):

PR #177, opened 2023-07-06
Feature/docs mermaid

andresharpe:
- Validating both sides of relationship exists when building Solution
- Added new Nox.Docs project and extension method to build Mermaid.js ERD
- Supports Summary, Normal and Detailed diagrams
- TODO: Owned entities and composite keys (implied relationships)


jan-schutte (approved):

rochar (approved):

rochar (approved):

src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -69,5 +70,103 @@ internal bool ApplyDefaults()
...
+
+            if (OwnedRelationships is not null)
+            {
+                foreach (var relationship in OwnedRelationships)
+                {
+                    if (relationship.Related.Entity?.Keys is null)
+                    {
+                        if (relationship.Relationship == EntityRelationshipType.OneOrMany || relationship.Relationship == EntityRelationshipType.ZeroOrMany)
+                        {
+                            var foreignKeyName = $"{relationship.Related.Entity!.PluralName}";
```
alexandervlasenko, line 97: The code blocks inside "if" and "else" are practically identical - maybe it makes sense to have only foreignKeyName and Type logic inside if/else block to highlight the actual difference.

src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -69,5 +70,103 @@ internal bool ApplyDefaults()
...
+                                Type = Types.NoxType.AutoNumber,
+                                IsRequired = true,
+                                IsReadonly = true,
+                            };
+                            yield return new(EntityMemberType.OwnedRelationship, foreignKey);
+                        }
+                    }
+                    else
+                    {
+                        foreach (var key in relationship.Related.Entity.Keys)
```
alexandervlasenko, line 124: The code is practically the same as in relationships below (lines 140-145) except setting the type (EntityMemberType.OwnedRelationship/EntityMemberType.Relationship), so maybe it is possible to extract a common method.

src/Nox.Docs/Extensions/NoxSolutionMermaidExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+
+                    sb.AppendLine($"        {memberType} {memberName}{keyType}{description}");
+                }
+            }
+
+            sb.AppendLine($"    }}");
+
+            if (entity.Relationships is not null)
+            {
+                foreach (var relationship in entity.Relationships)
```
alexandervlasenko, line 55: Should we also have OwnedRelationships here?

PR #181, opened 2023-07-07
Make keys mandatory for entities

ILoveMyTortoise:


ILoveMyTortoise:
image
image

jan-schutte (approved):

rochar (changes requested):
nice.

Can you add a test to YamlFileValidationTests, with an incorrect yaml (no keys)

alexandervlasenko (dismissed):

ILoveMyTortoise:
> nice. Can you add a test to YamlFileValidationTests, with an incorrect yaml (no keys)

@rochar Actually, we already have a test with complex validation of error scenarios, Deserialize_WithNoxYamlSerializer_ThrowsException, so I've increased the error count. Maybe in the future we can add check for particular text error messages

alexandervlasenko (approved):

jan-schutte (approved):

rochar (approved):

PR #192, opened 2023-07-09
NoxUri Implementation added

fkucuk:
System.Uri is sufficient for us. Just marked as ValueObject


src/Nox.Types/Types/NoxUri/NoxUri.cs
```diff
@@ -0,0 +1,4 @@
+namespace Nox.Types;
+public sealed class NoxUri : ValueObject<System.Uri, NoxUri> 
+{
```
rochar, line 3: would be also benefit to have the From(string) construction

src/Nox.Types/Types/NoxUri/NoxUri.cs
rochar, line 1: Can call it Uri, no need to have Nox, like other types (the same for the folder)

src/Nox.Types/Types/NoxUri/NoxUri.cs
rochar: Consider also the UriKindType

src/Nox.Types/Types/NoxUri/NoxUri.cs
```diff
@@ -0,0 +1,4 @@
+namespace Nox.Types;
+public sealed class NoxUri : ValueObject<System.Uri, NoxUri>
```
rochar, line 2: Uri keeps the URI and the UriKind, so this will be a Compound Type

tests/Nox.Types.Tests/Types/NoxUri/NoxUriTests
```diff
@@ -0,0 +1,12 @@
...
+public class NoxUriTests
+{
+    [Fact]
+    public void NoxUri_ShouldBe_SameAs_SystemUri()
+    {
+        var expected = new Uri("http://www.regus.com");
+        var actual = NoxUri.From(expected);
+        actual.Value.Should().Be(expected);
+    }
+}
```
rochar, line 12: the NoxTypesEntityFrameworkTests is not very well designed, one test does all, but you should have a test for this new type and validate how the EF handles it, and that no converter is needed

rochar (changes requested):
looks good, a few comments

in the Customer Perspective (Dev Team we may think what kind of feature would be beneficial to have on this type, so its no so anemic)

src/Nox.Types/Types/NoxUri/NoxUri.cs
fkucuk: To avoid ambiguity, I'd preferred to call it that way. But you are right. I've changed it.

tests/Nox.Types.Tests/Types/NoxUri/NoxUriTests.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Security.Policy;
+using FluentAssertions;
+using Xunit.Abstractions;
+
+namespace Nox.Types.Tests.Types;
+
+public class UriTests
+{
+
+    private const string SAMPLE_URI = "https://user:password@www.contoso.com:80/Home/Index.htm?q1=v1&q2=v2#FragmentName";
```
rochar, line 11: preferer Sample_Uri instead

tests/Nox.Types.Tests/Types/NoxUri/NoxUriTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    private readonly ITestOutputHelper _testOutputHelper;
+
+
+    public UriTests(ITestOutputHelper testOutputHelper)
+    {
+        _testOutputHelper = testOutputHelper;
+    }
+
+    [Fact]
+    public void NoxUri_ShouldBe_SameAs_SystemUri()
```
rochar, line 21: Can you remove the Nox Prefix?

src/Nox.Types/Types/Uri/Uri.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Collections.Generic;
+
+namespace Nox.Types;
+
+public sealed class Uri : ValueObject<System.Uri, Uri> 
+{
+
+    public static Uri From(string value, System.UriKind uriKind = System.UriKind.Absolute)
```
rochar, line 9: I understand that the URI per si already does the validation, so we need to do this in the constructor

rochar (approved):
I agree for this first iteration, no need to have a type options, lets see the concrete use cases

tests/Nox.Types.Tests/Types/NoxUri/NoxUriTests.cs
fkucuk: I think this can stay, used in Test Method name to differentiate between System.Uri

tests/Nox.Types.Tests/Types/NoxUri/NoxUriTests.cs
fkucuk: done

tests/Nox.Types.Tests/Types/NoxUri/NoxUriTests
fkucuk: done

src/Nox.Types/Types/NoxUri/NoxUri.cs
fkucuk: System.Uri already has this as a constructor. 
I believe current implementation meets our needs.

jan-schutte (approved):
Nice job Fatih

alexandervlasenko (approved):

rochar:
@fkucuk update to the main branch state and y can merge this PR thanks

PR #198, opened 2023-07-10
Implemented Date type #101

emir-sehovic1:


src/Nox.Types/Types/Date/Date.cs
```diff
@@ -0,0 +1,149 @@
...
+        => ToString(format, CultureInfo.InvariantCulture);
+
+    public string ToString(IFormatProvider formatProvider)
+        => ToString("d", formatProvider);
+
+    public string ToString(string format, IFormatProvider formatProvider)
+    {
+        if (!IsValidDateOnlyFormat(format))
+        {
+            throw new FormatException("Input string was not in a correct format.");
```
dejobratic, line 113: This message is fine. I'm just wondering if maybe something like "The provided format is not supported." would be more descriptive?

src/Nox.Types/Types/Date/Date.cs
```diff
@@ -0,0 +1,149 @@
...
+            result.Errors.Add(new ValidationFailure(nameof(Value), $"Could not create a Nox Number type as value {Value.Date.ToString("d", CultureInfo.InvariantCulture)} is greater than than the maximum specified value of {_dateTypeOptions.MaxValue.Date.ToString("d", CultureInfo.InvariantCulture)}"));
+        }
+
+        return result;
+    }
+
+    /// <summary>
+    /// Gets the year of the date.
+    /// </summary>
+    public int Year => Value.Year;
```
dejobratic, line 78: We need to check whether the public properties should be marked as NotMapped.

dejobratic (commented):
Good job. Left some minor comments to discuss.

dejobratic (commented):
Looks good.

src/Nox.Types/Types/Date/Date.cs
emir-sehovic1: If custom converter is not used, then only the properties with public setters would be persisted

dejobratic (approved):
Good job

rochar (approved):
looks good

PR #200, opened 2023-07-10
Nox generator/number type update

AndreyDegtyarev:
Use QuantityValue in Number type.


rochar (approved):
look good

jan-schutte (commented):
Look good, just the question about underlying type

jan-schutte (approved):

PR #201, opened 2023-07-10
Feature/add file content tests

ILoveMyTortoise:


rochar (approved):
looks good, not sure if we went to far, time will tell :)

just review my single comment

PR #203, opened 2023-07-10
Feature Boolean Type

danield-ap:


gjorgji-grgovski-authoritypartners (approved):

src/Nox.Types/Types/Boolean/Boolean.cs
rochar, line 1: We need also the converter for EF, and add the test to this type on EF.Tests

rochar (changes requested):
looks good, just need to have a converter

src/Nox.Types/Types/Boolean/Boolean.cs
danield-ap: Converter created and boolean property added to EF.Tests

PR #208, opened 2023-07-11
Added ToMarkdown method to NoxSolutionMermaidExtensions and added cal…

Mornevanzyl:
- Added ToMarkdown() method to NoxSolutionMermaidExtensions class
- Added call from tests to output .md files to same folder as source solution yaml.

### DIFF
--- a/tests/Nox.Docs.Tests/Nox.Docs.Tests/NoxDocsMermaidTests.cs
+++ b/tests/Nox.Docs.Tests/Nox.Docs.Tests/NoxDocsMermaidTests.cs
@@ -14,6 +14,8 @@ public class NoxDocsMermaidTests
 
         var mermaidText = noxSolution.ToMermaidErd();
 
+        noxSolution.ToMarkdown(mermaidText, $"./files/{noxSolution.Name}");
+
         var expectedResult = """
             erDiagram
                 Country {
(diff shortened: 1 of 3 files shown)


tests/Nox.Docs.Tests/Nox.Docs.Tests/NoxDocsMermaidTests.cs
```diff
@@ -14,6 +14,8 @@ public void Solution_Creates_Valid_Mermaid_Erd()
 
         var mermaidText = noxSolution.ToMermaidErd();
 
+        noxSolution.ToMarkdown(mermaidText, $"./files/{noxSolution.Name}");
```
rochar, line 17: Would make things more clear, if we have a test just "Generate_MarkDown"

tests/Nox.Docs.Tests/Nox.Docs.Tests/NoxDocsMermaidTests.cs
```diff
@@ -66,6 +68,8 @@ public void Solution_Creates_Valid_Mermaid_DetailedErd()
 
         var mermaidText = noxSolution.ToMermaidErd(ErdDetail.Detailed);
 
+        noxSolution.ToMarkdown(mermaidText, $"./files/{noxSolution.Name}_{ErdDetail.Detailed}");
```
rochar, line 71: The Same here

rochar (changes requested):

PR #210, opened 2023-07-11
MP-7420: Time definition and implementation

originalethanhunt:


src/Nox.Types/Types/Time/Time.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
+using System.Globalization;
+
+namespace Nox.Types;
+
+/// <summary>
+/// Represents a time of day, as would be read from a clock, within the range 00:00:00 to 23:59:59.9999999.
+/// </summary>
+public class Time : ValueObject<DateTime, Time> // This should be TimeOnly
```
dejobratic, line 9: Can we mark the class as sealed?

src/Nox.Types/Types/Time/Time.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    /// <summary>
+    /// MinTimeTicks is the ticks for the midnight time 00:00:00.000 AM
+    /// </summary>
+    private static long MinTimeTicks = 0;
```
dejobratic, line 14: Can we mark the field as constant?

src/Nox.Types/Types/Time/Time.cs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// <summary>
+    /// MaxTimeTicks is the max tick value for the time in the day. It is calculated using DateTime.Today.AddTicks(-1).TimeOfDay.Ticks.
+    /// </summary>
+    private static long MaxTimeTicks = 863_999_999_999;
```
dejobratic, line 19: Can we mark the field as constant?

src/Nox.Types/Types/Time/Time.cs
```diff
@@ -0,0 +1,100 @@
...
+            throw new TypeValidationException(result.Errors);
+        }
+
+        newObject.Value = new DateTime().Add(timeSpan);
+
+        return newObject;
+
+    }
+
+    public override string ToString() => ToString(CultureInfo.InvariantCulture);
```
dejobratic, line 91: Can we add comments to ToString methods, since they are public?

dejobratic (changes requested):
Good job. Just a few small comments.

dejobratic (approved):

src/Nox.Types/Types/Time/Time.cs
```diff
@@ -0,0 +1,124 @@
+﻿using System;
+using System.Globalization;
+
+namespace Nox.Types;
+
+/// <summary>
+/// Represents a time of day, as would be read from a clock, within the range 00:00:00 to 23:59:59.9999999.
+/// </summary>
+public sealed class Time : ValueObject<DateTime, Time> // This should be TimeOnly
```
rochar, line 9: is there a reason to not be TimeOnly Struct? for the underline type?

src/Nox.Types/Types/Time/Time.cs
```diff
@@ -0,0 +1,124 @@
...
+
+    }
+
+    private static Time From(TimeSpan timeSpan)
+    {
+        var newObject = new Time();
+
+        var result = newObject.Validate();
+        
+        if (timeSpan.Ticks < Time.MinTimeTicks)
```
rochar, line 71: Should this be done on the Validate method?

rochar (changes requested):

src/Nox.Types/Types/Time/Time.cs
rochar, line 1: I would expect a TimeTypeOptions, defining min and max for example

rochar (changes requested):

src/Nox.Types/Types/Time/Time.cs
originalethanhunt: TimeOnly is not available as part of .Net standard 2.0 framework.

src/Nox.Types/Types/Time/Time.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            result.Errors.Add(new ValidationFailure(nameof(Value), $"Could not create a Nox Time type as value {timeSpan} is greater than than the maximum specified value of {MaxTimeTicks}."));
+        }
+
+        if (!result.IsValid)
+        {
+            throw new TypeValidationException(result.Errors);
+        }
+
+        newObject.Value = new DateTime().Add(timeSpan);
```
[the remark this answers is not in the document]
originalethanhunt: TimeSpan was used as it worked well with negative tick count and negative time of the day, which resulted in ArgumentOutOfRange exception with DateTime before the invocation of Validate method. However I have updated the implementation addressing your concerns.

src/Nox.Types/Types/Time/Time.cs
originalethanhunt: Added

src/Nox.Types/Types/Time/Time.cs
originalethanhunt: Modified the implementation, this was a validation on intermediate timespan and not on the TValue(DateTime).

src/Nox.Types/Types/Time/TimeTypeOptions.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+namespace Nox.Types;
+
+public class TimeTypeOptions
+{
+    public long MinTime { get; internal set; } = 0;
```
rochar, line 6: Its not clear what this represents, Ticks?

src/Nox.Types/Types/Time/Time.cs
rochar: Can you copy this class and see if it compiles without an extra depdendency https://github.com/dotnet/dotnet/blob/5b09052ac0948cc1a94829f01491c550d4837757/src/runtime/src/libraries/System.Private.CoreLib/src/System/TimeOnly.cs#L4 ?

I'm taking care of migrating our pacakages to a more recent version

src/Nox.Types/Types/Time/TimeTypeOptions.cs
rochar, line 1: We need to add a Field into NoxSimpleTypeDefinition for this type, see there other examples
Thanks

rochar (changes requested):
thanks, just a few comments

rochar (changes requested):
just a small comment

src/Nox.Types.Abstractions/TypeDefinitions/NoxSimpleTypeDefinition.cs
```diff
@@ -61,6 +61,9 @@ public class NoxSimpleTypeDefinition
     [IfEquals("Type", NoxType.DateTime)]
     public DateTimeTypeOptions? DateTimeTypeOptions { get; set; }
 
+    [IfEquals("Type", NoxType.Time)]
```
rochar, line 64: you need to regenerate the schema, use this test NoxSolutionSchemaGenerate and commit the changes

rochar (changes requested):

src/Nox.Types.Abstractions/TypeDefinitions/NoxSimpleTypeDefinition.cs
originalethanhunt: Updated

PR #216, opened 2023-07-12
Refactored measurement types

dejobratic:
- All implemented measurement types (Area, Length, Distance and Volume) now inherit Measurement base class;
- Added relevant units and conversion factors;
- Changed tests to FluentAssertions;


src/Nox.Types/Types/Area/Area.cs
```diff
@@ -80,47 +49,33 @@ internal override ValidationResult Validate()
     {
         var result = Value.Validate();
 
-        if (Value < 0)
+        if (!double.IsNaN((double)Value) && !double.IsInfinity((double)Value))
```
rochar, line 52: if IsNan or infinity should we also add an error for this case?

rochar (changes requested):
looks good, just a question

rochar (approved):

PR #217, opened 2023-07-12
Implemented Date Time Duration type #99

emir-sehovic1:
- Created a Date Time Duration value object with TimeSpan as underlying type
- Date Time Duration can be created from native TimeSpan type or from various options by specifying days, hours, minutes, seconds or milliseconds


tests/Nox.Types.Tests/Types/DateTimeDuration/DateTimeDurationTests.cs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    [Fact]
+    public void From_WithNegativeTimeSpanInput_ReturnsAbsoluteValue()
+    {
+        var timeSpan = new TimeSpan(1, 2, 3, 4, 5);
+
+        var dateTimeDuration = DateTimeDuration.From(-timeSpan);
+
+        dateTimeDuration.Value.Should().Be(timeSpan.Duration());
```
dejobratic, line 26: Can we have a hardcoded expected value instead of timeSpan.Duration()?

dejobratic (changes requested):
Good job. Just a minor comments on tests.

tests/Nox.Types.Tests/Types/DateTimeDuration/DateTimeDurationTests.cs
emir-sehovic1: Thanks for the suggestion, I've updated the tests

dejobratic (approved):

src/Nox.Types/Types/DateTimeDuration/DateTimeDuration.cs
rochar, line 1: We need to think if we need a TypesOptions, for example Max/Min Duration

rochar (changes requested):
looks good small comment

rochar:
please also update to main

src/Nox.Types/Types/DateTimeDuration/DateTimeDuration.cs
emir-sehovic1: Thanks for the suggestion, I've added the duration options

src/Nox.Types/Types/DateTimeDuration/DateTimeDurationTypeOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+    public static readonly int DefaultMinDurationMinutes = 0;
+    public static readonly int DefaultMinDurationSeconds = 0;
+    public static readonly int DefaultMinDurationMilliseconds = 0;
+
+    public int MaxDurationDays { get; set; } = DefaultMaxDurationDays;
+    public int MaxDurationHours { get; set; } = DefaultMaxDurationHours;
+    public int MaxDurationMinutes { get; set; } = DefaultMaxDurationMinutes;
+    public int MaxDurationSeconds { get; set; } = DefaultMaxDurationSeconds;
+    public int MaxDurationMilliseconds { get; set; } = DefaultMaxDurationMilliseconds;
+    public TimeSpan GetMaxDuration() => new TimeSpan(MaxDurationDays, MaxDurationHours, MaxDurationMinutes, MaxDurationSeconds, MaxDurationMilliseconds).Duration();
```
rochar, line 24: TypeOptions looks good. My expectation to GetMaxDuration was to use a single field to define the max timespan, for example TimeSpan.FromMilliseconds()... From a user perspective I would argue that the expectation would be to create a MaxTimeSpan from a unit... and not set all this values, to mee seems complex..

rochar (changes requested):
looks good, just a small topic to discuss

rochar (approved):

rochar:
fix the conflicts, update with main and you can merge. Thks

PR #220, opened 2023-07-12
Features/entity key is required validation

AndreyDegtyarev:
Added custom JsonConverter to deserialize List<Entity> in order to validate specific rules.


rochar (approved):
I think this solution is more simple using the validator

jan-schutte (approved):

PR #225, opened 2023-07-13
Implemented Weight type

dejobratic:
- Implemented Weight type that inherits Measurement;
- Current supported units are kilogram and pound;

### DIFF
--- a/src/Nox.Types/Types/Area/Area.cs
+++ b/src/Nox.Types/Types/Area/Area.cs
@@ -6,7 +6,7 @@ namespace Nox.Types;
 /// <summary>
 /// Represents a Nox <see cref="Area"/> type and value object.
 /// </summary>
-public class Area : Measurement<Area, AreaUnit>
+public sealed class Area : Measurement<Area, AreaUnit>
 {
     private const long EarthsSurfaceAreaInSquareMeters = 510_072_000_000_000;
 
--- a/src/Nox.Types/Types/Weight/Weight.cs
+++ b/src/Nox.Types/Types/Weight/Weight.cs
@@ -1,11 +1,51 @@
 using Nox.Types.Common;
+using System.Collections.Generic;
 
 namespace Nox.Types;
 
+/// <summary>
+/// Represents a Nox <see cref="Weight"/> type and value object.
+/// </summary>
+public sealed class Weight : Measurement<Weight, WeightUnit>
+{
     /// <summary>
-    /// Represents a Nox <see cref="Weight"/> type and value object.
+    /// Creates a new instance of <see cref="Weight"/> object in kilograms.
     /// </summary>
-    /// <remarks>Placeholder, needs to be implemented</remarks>
-    public sealed class Weight : ValueObject<QuantityValue, Weight>
+    /// <param name="value">The value to create the <see cref="Weight"/> with</param>
+    /// <returns></returns>
+    /// <exception cref="TypeValidationException"></exception>
+    public static Weight FromKilograms(QuantityValue value)
+        => From(value, WeightUnit.Kilogram);
+
+    /// <summary>
+    /// Creates a new instance of <see cref="Weight"/> object in pounds.
+    /// </summary>
+    /// <param name="value">The origin value to create the <see cref="Weight"/> with</param>
+    /// <returns></returns>
+    /// <exception cref="TypeValidationException"></exception>
+    public static Weight FromPounds(QuantityValue value)
+        => From(value, WeightUnit.Pound);
+
+    /// <summary>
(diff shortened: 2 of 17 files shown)


tests/Nox.Types.Tests/Types/Weight/WeightTests.cs
```diff
@@ -1,11 +1,219 @@
...
+
+        actual.Equals(expected).Should().BeTrue();
+
+        (expected == actual).Should().BeTrue();
+
+        (expected != actual).Should().BeFalse();
+    }
+
+    private static void AssertAreNotEquivalent(Weight expected, Weight actual)
+    {
```
rochar, line 208: You can/should use AssertionScope for this

src/Nox.Types/Types/Weight/Weight.cs
rochar, line 1: @dejobratic  we will need the database converter for simple types, and the EF Test to validate that is doing what we expect

Also consider to have a TypeOptions to define a max and min at leats

rochar (changes requested):
Looks good, just two notes that we learned meanwhile and need to be applied to all Measurements

src/Nox.Types/Types/Area/Area.cs
```diff
@@ -6,7 +6,7 @@ namespace Nox.Types;
 /// <summary>
 /// Represents a Nox <see cref="Area"/> type and value object.
 /// </summary>
-public class Area : Measurement<Area, AreaUnit>
+public sealed class Area : Measurement<Area, AreaUnit>
```
rochar, line 9: ok.

We already saw that probably we will need to change this because o EF LayzLoadProxy requirements

tests/Nox.Types.Tests/Types/Weight/WeightTests.cs
dejobratic: Thanks for the suggestion. Added.

src/Nox.Types/Types/Weight/Weight.cs
dejobratic: Added database converters.

src/Nox.Types/Types/Area/Area.cs
dejobratic: Thanks for the info. Removed in all measurement types.

src/Nox.Types.EntityFramework/Types/Volume/VolumeConverter.cs
```diff
@@ -0,0 +1,12 @@
+﻿using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+
+namespace Nox.Types.EntityFramework.Types;
+
+public class VolumeToCubicFootConverter : ValueConverter<Volume, double>
```
gjorgji-grgovski-authoritypartners, line 5: Typo: VolumeToCubicFootConverter -> VolumeToCubicFeetConverter

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types.EntityFramework/Types/Volume/VolumeConverter.cs
dejobratic: Fixed.

gjorgji-grgovski-authoritypartners (approved):

src/Nox.Types/Types/Weight/Weight.cs
```diff
@@ -1,11 +1,51 @@
...
+    /// <summary>
+    /// Creates a new instance of <see cref="Weight"/> object in kilograms.
+    /// </summary>
+    /// <param name="value">The origin value to create the <see cref="Weight"/> with</param>
+    /// <returns></returns>
+    /// <exception cref="TypeValidationException"></exception>
+    public new static Weight From(QuantityValue value)
+        => From(value, WeightUnit.Kilogram);
+
+    private QuantityValue? _kilograms;
```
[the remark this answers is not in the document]
dejobratic: It acts as a simple memory cache for the value, so that we don't calculate the value each time.

src/Nox.Types/Types/Weight/Weight.cs
```diff
@@ -1,11 +1,51 @@
...
+    public QuantityValue ToKilograms() => _kilograms ??= GetMeasurementIn(WeightUnit.Kilogram);
+
+    private QuantityValue? _pounds;
```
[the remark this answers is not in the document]
dejobratic: It acts as a simple memory cache for the value, so that we don't calculate the value each time.

rochar (approved):

PR #229, opened 2023-07-13
Color

denisvieira-dev:


src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// <summary>
+    /// Validates a <see cref="Color"/> object.
+    /// </summary>
+    /// <returns>true if the <see cref="Color"/> value is valid.</returns>
+    internal override ValidationResult Validate()
+    {
+        var result = base.Validate();
+
+        if (System.Drawing.Color.FromArgb(Alpha, Red, Green, Blue).IsEmpty)
```
danield-ap, line 44: Checking in the [Color.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Drawing.Primitives/src/System/Drawing/Color.cs) code I couldn't see any way of `IsEmpty` ever being true, so it doesn't validate anything for us.
But `FromArgb` can throw System.ArgumentException in some situations, but since this is a validate method that adds the validations to out Error array, we should prevent it from throwing an exception.
I think we could do something like `try { System.Drawing.Color.FromArgb(Alpha, Red, Green, Blue) } catch ArgumentException e { result.Errors.Add(new ValidationFailure(e.Message); }` .

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,179 @@
+﻿using System.Linq;
+
+namespace Nox.Types;
+
+/// <summary>
+/// Represents a Nox <see cref="Color"/> type and value object. 
+/// </summary>
+public class Color : ValueObject<(byte Alpha, byte Red, byte Green, byte Blue), Color>
```
danield-ap, line 8: Have you considered storing color as a single value?
I don't think there's any having them like this if we won't use them separately.
It also adds to complexity when storing/retrieving these values that could be represented by a single number/string.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,179 @@
...
+    /// Creates a new instance of <see cref="Color"/> object
+    /// </summary>
+    /// <param name="value">The origin value to create the <see cref="Color"/> with</param>
+    /// <returns></returns>
+    /// <exception cref="TypeValidationException"></exception>
+    public static Color FromRgba(byte alpha, byte red, byte green, byte blue)
+    {
+        var color = System.Drawing.Color.FromArgb(alpha, red, green, blue);
+
+        var newObject = new Color
```
danield-ap, line 65: You have these exact same lines of code repeated in 3 methods:
```
var newObject = new Color
{
    Alpha = color.A,
    Red = color.R,
    Green = color.G,
    Blue = color.B,
};

var validationResult = newObject.Validate();

if (!validationResult.IsValid)
{
    throw new TypeValidationException(validationResult.Errors);
}

return newObject;
```

danield-ap (changes requested):

src/Nox.Types/Types/Color/Color.cs
gjorgji-grgovski-authoritypartners: The requirement was to persist it as 4 bytes.
Maybe "byte[]", but you have to check how it would be persisted in the database.

image

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,157 @@
...
+    /// Creates a new instance of <see cref="Color"/> object
+    /// </summary>
+    /// <param name="value">The origin value to create the <see cref="Color"/> with</param>
+    /// <returns></returns>
+    /// <exception cref="TypeValidationException"></exception>
+    public static Color FromAlphaColor(string alphadecimal)
+    {
+        return ColorConverter.ConvertFromString(alphadecimal, CultureInfo.InvariantCulture);
+    }
+
```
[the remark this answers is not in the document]
denisvieira-dev: Hello @gjorgji-grgovski-authoritypartners,  thanks for your feedback. I've taken note of your suggestion about creating a FromRgb method. However, the native C# library doesn't provide a FromRgb method for System.Drawing.Color.
As an alternative approach, I've set the Alpha value to 0 (which makes the color fully transparent) when we receive an RGB color. This ensures the Color instance is created with Alpha set to 0 when we only have RGB values.
On the other hand, when we receive an ARGB color, all parameters (including Alpha) are correctly populated. This implementation caters to both ARGB and RGB inputs while maintaining color information as accurately as possible.

src/Nox.Types/Types/Color/Color.cs
denisvieira-dev: Each element of the tuple should correspond to a column in a table. For this tuple representing RGBA color values, we can create a table with four columns (Red, Green, Blue, Alpha). and then store the tuple's data. Similarly, to retrieve the tuple, execute a SELECT and construct the tuple back from the query result.

src/Nox.Types/Types/Color/ColorConverter.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+
+namespace Nox.Types;
+
+public class ColorConverter
```
danield-ap, line 7: Make this class static since all its members are static.

src/Nox.Types/Types/Color/ColorConverter.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    public static Color ConvertFromString(string strValue, CultureInfo culture)
```
danield-ap, line 9: There's no unit test for this method.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,148 @@
...
+    {
+        var color = System.Drawing.Color.FromArgb(alpha, red, green, blue);
+
+        return CreateColor(color);
+    }
+
+    /// <summary>
+    /// Creates a new instance of <see cref="Color"/> object
+    /// </summary>
+    /// <param name="value">The origin value to create the <see cref="Color"/> with</param>
```
danield-ap, line 67: Param tag is not matching in the actual param name.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,148 @@
...
+    {
+        var color = System.Drawing.Color.FromName(name);
+
+        return CreateColor(color);
+    }
+
+    /// <summary>
+    /// Creates a new instance of <see cref="Color"/> object
+    /// </summary>
+    /// <param name="value">The origin value to create the <see cref="Color"/> with</param>
```
danield-ap, line 80: Param tag is not matching in the actual param name.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,148 @@
...
+    private byte _red { get; set; }
+    private byte _green { get; set; }
+    private byte _blue { get; set; }
+    
+    public Color()
+    {
+        Value = new byte[4];
+    }
+
+    private bool _isEmpty { get; set; }
```
danield-ap, line 22: I'm not sure what is this property for since it's never updated.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    public static Color Empty
```
danield-ap, line 24: There should be unit tests for this property. 
Suggestions:
```c#
        new Color().Should().Be(Color.Empty);
        Color.FromRgba(0, 0, 0, 0).Should().Be(Color.Empty);
        Color.Empty.Should().Be(Color.Empty);
        Color.From(255, 255, 255).Should().NotBe(Color.Empty);
        ColorConverter.ConvertFromString("", CultureInfo.InvariantCulture).Should().Be(Color.Empty);
```

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        if (Value.Length > 4)
+            result.Errors.Add(new ValidationFailure(nameof(Value),
+                $"Could not create a Nox {nameof(Color)} type with more than 4 values."));
+
+        return result;
+    }
+
+    public static Color From(byte red, byte green, byte blue)
+        => From(new byte[] { 0, red, green, blue });
```
danield-ap, line 49: I think we should set Alpha for the maximum value. It doesn't make sense for it to return me a transparent (no color) when I'm creating a colors by sending its RGB values.
This is a just a suggestion since it's more of a design choice.

src/Nox.Types/Types/Color/Color.cs
danield-ap, line 1: We should have xml comment for every public method.

tests/Nox.Types.Tests/Types/Color/ColorTests.cs
danield-ap, line 1: Add a test for equality like
`Color.From(1, 2, 3).Should().Be(Color.From(1, 2, 3));`

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,148 @@
...
+
+    /// <summary>
+    /// Creates a new instance of <see cref="Color"/> object
+    /// </summary>
+    /// <param name="value">The origin value to create the <see cref="Color"/> with</param>
+    /// <returns></returns>
+    /// <exception cref="TypeValidationException"></exception>
+    public static Color FromRgba(byte alpha, byte red, byte green, byte blue)
```
danield-ap, line 57: Parameters are not included in the XML comment.

danield-ap (changes requested):

src/Nox.Types/Types/Color/Color.cs
denisvieira-dev: Its raised from converter, if the converter wasn't capable of create a new Color type.

src/Nox.Types/Types/Color/Color.cs
denisvieira-dev: I think it could be ok, because 255 in alpha should be fully opaque.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,213 @@
...
+    /// </summary>
+    public Color()
+    {
+        Value = new byte[4];
+    }
+
+    /// <summary>
+    /// Creates a new instance of <see cref="Color"/> object with empty color.
+    /// </summary>
+    public static Color Empty
```
gjorgji-grgovski-authoritypartners, line 28: We do not need this static property, because we can always instantiate an empty color by:
```
var color = new Nox.Types.Color();
or
var emptyColor = new Nox.Types.Color().From(System.Drawing.Color.Empty)
```

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,213 @@
...
+
+    /// <summary>
+    /// Validates a <see cref="Color"/> object.
+    /// </summary>
+    /// <returns>true if the <see cref="Color"/> value is valid.</returns>
+    internal override ValidationResult Validate()
+    {
+        var result = new ValidationResult();
+
+        if (Value.Length < 1)
```
gjorgji-grgovski-authoritypartners, line 41: This is impossible to happen

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,213 @@
...
+            result.Errors.Add(new ValidationFailure(nameof(Value),
+                $"Could not create a Nox {nameof(Color)} type as an empty {nameof(Value)} is not allowed."));
+
+        if (Value.Length > 4)
```
gjorgji-grgovski-authoritypartners, line 45: This is impossible to happen

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,213 @@
...
+
+    /// <summary>
+    /// Creates a new instance of <see cref="Color"/> object.
+    /// </summary>
+    /// <param name="alpha">The alpha color.</param>
+    /// <param name="red">The red color.</param>
+    /// <param name="green">The green color.</param>
+    /// <param name="blue">The blue color.</param>
+    /// <exception cref="TypeValidationException"></exception>
+    public static Color FromRgba(byte alpha, byte red, byte green, byte blue)
```
gjorgji-grgovski-authoritypartners, line 74: Rename the method from "FromRgba" to "From"

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,213 @@
...
+    {
+        var color = System.Drawing.Color.FromArgb(alpha, red, green, blue);
+
+        return CreateColor(color);
+    }
+
```
gjorgji-grgovski-authoritypartners, line 80: Can we have a "From" method overloading which accepts System.Drawing.Color?

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,213 @@
...
+    {
+        var color = System.Drawing.Color.FromArgb(_alpha, _red, _green, _blue);
+
+        var knownColor = KnownColor.Colors.FirstOrDefault(kc => kc.Value.ToArgb() == color.ToArgb());
+
+        return knownColor.Key ?? "UnknownColor";
+    }
+
+    /// <inheritdoc />
+    public override string ToString()
```
gjorgji-grgovski-authoritypartners, line 137: We do not need this overriding of the method as it should already work like that with the Base implementation, please add a test for this.

src/Nox.Types/Types/Color/ColorConverter.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+
+namespace Nox.Types;
+
+public static class ColorConverter
```
gjorgji-grgovski-authoritypartners, line 7: Can we rename this one to Helper or something like that, because we are using Converters for the EnityFramework?

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types/Types/Color/ColorConverter.cs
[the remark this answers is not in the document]
denisvieira-dev: Done

src/Nox.Types/Types/Color/Color.cs
denisvieira-dev: Done

src/Nox.Types/Types/Color/Color.cs
denisvieira-dev: Removed

src/Nox.Types/Types/Color/Color.cs
denisvieira-dev: Done

src/Nox.Types/Types/Color/Color.cs
denisvieira-dev: Done

src/Nox.Types/Types/Color/ColorConverter.cs
denisvieira-dev: Done

src/Nox.Types/Types/Color/ColorHelper.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+
+namespace Nox.Types;
+
+public static class ColorHelper
```
gjorgji-grgovski-authoritypartners, line 7: Please add tests for this helper class.

gjorgji-grgovski-authoritypartners (approved):

src/Nox.Types/Types/Color/KnownColor.cs
```diff
@@ -0,0 +1,164 @@
...
+        { "Turquoise", System.Drawing.Color.Turquoise },
+        { "Violet", System.Drawing.Color.Violet },
+        { "Wheat", System.Drawing.Color.Wheat },
+        { "White", System.Drawing.Color.White },
+        { "WhiteSmoke", System.Drawing.Color.WhiteSmoke },
+        { "Yellow", System.Drawing.Color.Yellow },
+        { "YellowGreen", System.Drawing.Color.YellowGreen },
+    };
+
+    public static bool TryGetNamedColor(string text,  out System.Drawing.Color color)
```
rochar, line 149: this can be replace by this: `Enum.TryParse<System.Drawing.Color>(text, out color);`

src/Nox.Types/Types/Color/ColorHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+        return tokens.Length switch
+        {
+            1 => PossibleKnownColor(System.Drawing.Color.FromArgb(IntFromString(tokens[0], culture))),
+            3 => PossibleKnownColor(System.Drawing.Color.FromArgb(IntFromString(tokens[0], culture), IntFromString(tokens[1], culture), IntFromString(tokens[2], culture))),
+            4 => PossibleKnownColor(System.Drawing.Color.FromArgb(IntFromString(tokens[0], culture), IntFromString(tokens[1], culture), IntFromString(tokens[2], culture), IntFromString(tokens[3], culture))),
+            _ => new Nox.Types.Color(),
+        };
+    }
+
+    private static Color PossibleKnownColor(System.Drawing.Color color)
```
rochar, line 69: Why do we need the if, the return is the same, ?

`return Color.From(color.A, color.R, color.G, color.B);`

src/Nox.Types/Types/Color/KnownColor.cs
```diff
@@ -0,0 +1,164 @@
+﻿using System.Collections.Generic;
+
+namespace Nox.Types;
+
+public static class KnownColor
+{
+    public static Dictionary<string, System.Drawing.Color> Colors = new Dictionary<string, System.Drawing.Color>()
```
rochar, line 7: I think we can drop this dictionary, because this seems redundant

src/Nox.Types/Types/Color/KnownColor.cs
rochar: Promote this methodo to ColorHelper

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -0,0 +1,195 @@
...
+    /// Returns the string representation of the color value, formatted in Hex format string
+    /// </summary>
+    /// <returns>The string representation</returns>
+    public string ToHex() => $"#{_red:X2}{_green:X2}{_blue:X2}";
+
+    /// <summary>
+    /// Returns the string representation of the color value, formatted in name format string
+    /// </summary>
+    /// <returns>The string representation</returns>
+    public string ToName()
```
rochar, line 109: for nox.types we want to keep simple as possible, so just expose the color:

return the color = System.Drawing.Color.FromArgb(_alpha, _red, _green, _blue);

The user can use this to get the name and whatever isn needed

rochar (changes requested):
looks good, lets just simplify

rochar (approved):

PR #233, opened 2023-07-14
Nox types/nuid type introduce

AndreyDegtyarev:


src/Nox.Types/Common/Base36Converter.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Text;
+
+namespace Nox.Types.Common;
+
+/// <summary>
+/// Utility methods for converting value to Base36 string.
+/// </summary>
+public static class Base36Converter
```
rochar, line 8: we can make this internal?

tests/Nox.Types.Tests/Types/Nuid/NuidTests.cs
```diff
@@ -4,8 +4,16 @@ namespace Nox.Types.Tests.Types;
...
     [Fact]
-    public void When_Create_Should()
+    public void StringValue_CreateManyInstances_ShouldBeEqual()
     {
+        const string testValue = "!#123TestValue456!#";
 
+        var nuidLeft = Nuid.From(testValue);
+        var nuidRight = Nuid.From(testValue);
+
+        Assert.Equal(nuidLeft, nuidRight);
```
rochar, line 14: Can you use FluentAssertions for this?

tests/Nox.Types.Tests/Types/Nuid/NuidTests.cs
```diff
@@ -4,8 +4,16 @@ namespace Nox.Types.Tests.Types;
 public class NuidTests
 {
```
rochar, line 5: A test ensuring that for a known string we have a known expected nuid its important, so no one breaks implementation

src/Nox.Types/Types/Nuid/NuidTypeOptions.cs
rochar, line 1: please also add this TypeOptions to NoxSimpleTypeDefinition, as the other types options

src/Nox.Types/Types/Nuid/Nuid.cs
rochar, line 1: simple types need to have a converter and we need to test it in Nox.Types.EF.Tests

rochar (changes requested):

src/Nox.Types/Types/Nuid/Nuid.cs
AndreyDegtyarev: This might be further changes. For now, there are no converters for any options.

tests/Nox.Types.Tests/Types/Nuid/NuidTests.cs
```diff
@@ -1,11 +1,31 @@
...
+    private const string TestStringValue = "!#123TestValue456!#";
+    private const uint ExpectedNuidValue = 598674021;
+    private const string ExpectedBase36 = "XTNFW9";
+    private readonly Guid ExpectedGuid = new("00000000-0000-0000-0000-000023af0a65");
+
     [Fact]
-    public void When_Create_Should()
+    public void CreateNuiidFromString_CheckImplementation_ShouldNotBeChanged()
```
rochar, line 14:
```suggestion
    public void FromString_Construction_ShouldReturnExpected()
```

src/SampleWebApp/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/AuditableEntityBase.g.cs
rochar, line 1: No change was done to affect the generated class, can y pleas check what is going one

tests/Nox.Types.Tests/Types/Nuid/NuidTests.cs
rochar: IMO this test implementation is not what we want to validate, because if we introduce an issue, this test will always be valid

rochar (changes requested):

rochar (approved):

alexandervlasenko (approved):

PR #252, opened 2023-07-17
Moved .sln File to Upper Folder

fkucuk:
To comply with the standards, I've moved the sln file to an upper folder.


rochar (changes requested):
Thanks, y need to review our Github pipelines

andresharpe:
I suggest we do this at the end. We will need to rename the repo and the solution file once we promote it to the new "Lib"

fkucuk:
nothing urgent, parked for later

PR #256, opened 2023-07-17
Implemeted file type #116

emir-sehovic1:
- Implemented file type that can be created from Url, PrettyName and SizeInBytes
- Supported file formats can be configured in options


src/Nox.Types/Types/File/File.cs
```diff
@@ -1,9 +1,129 @@
...
+
+    /// <summary>
+    /// Validates the <see cref="File"/> object.
+    /// </summary>
+    /// <returns>A validation result indicating whether the <see cref="File"/> object is valid or not.</returns>
+    internal override ValidationResult Validate()
+    {
+        var result = base.Validate();
+
+        System.Uri.TryCreate(Value.Url, UriKind.Absolute, out var x);
```
dejobratic, line 107: What is this statement used for?

dejobratic (commented):
Looks good. Just a small comment.

src/Nox.Types/Types/File/File.cs
emir-sehovic1: That was left in by mistake. Removed it now, thanks

dejobratic (approved):
Good job.

src/Nox.Types.Abstractions/Enums/File/FileFormatType.cs
```diff
@@ -0,0 +1,95 @@
...
+
+/// <summary>
+/// Enumeration for file format types.
+/// </summary>
+public enum FileFormatType
+{
+    /// <summary>
+    /// Represents all file format types.
+    /// </summary>
+    All,
```
rochar, line 15: to have this the enum must be decorated with Flags, and the All must be All = pdf | ....

rochar (changes requested):
Looks good just a small comment

rochar (approved):

PR #257, opened 2023-07-17
Feature/yaml value object

msadikkoseaupa:


tests/Nox.Types.Tests/EntityFrameworkTests/NoxTypesEntityFrameworkTests.cs
```diff
@@ -52,16 +52,17 @@ public void Countries_CanRead_LatLong()
             Uri = Uri.From(Sample_Uri),
             IsLandLocked = Boolean.From(false),
             DateTimeDuration = DateTimeDuration.From(days: 10, 5, 2, 1),
+            Yaml = Yaml.From("foo: bar"),
```
dejobratic, line 55: Can we provide a more meaningful yaml sample, like we did with Json here https://github.com/NoxOrg/Nox.Generator/blob/main/tests/Nox.Types.Tests/EntityFrameworkTests/NoxTypesEntityFrameworkTests.cs?

gjorgji-grgovski-authoritypartners (approved):

dejobratic (changes requested):
Good job. Just a minor comment.

tests/Nox.Types.Tests/EntityFrameworkTests/NoxTypesEntityFrameworkTests.cs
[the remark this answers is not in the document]
msadikkoseaupa: Yaml sample was updated with Switzerland's cities and counties

src/Nox.Types/Types/Yaml/Yaml.cs
```diff
@@ -0,0 +1,254 @@
...
+    private readonly List<YamlNode> _yamlNodes = new();
+    private bool _isValidated;
+
+    /// <summary>
+    /// Gets the equality components used for value comparison.
+    /// </summary>
+    /// <returns>The equality components.</returns>
+    protected override IEnumerable<KeyValuePair<string, object>> GetEqualityComponents()
+    {
+        var properties = GenerateKeyValuePairs();
```
rochar, line 23: you can return this method directly no need to iterate

rochar (changes requested):

src/Nox.Types/Types/Yaml/Yaml.cs
msadikkoseaupa: Method was updated to return generated properties directly.

src/Nox.Types/Types/Yaml/Yaml.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Validates the YAML object and returns the validation result.
+    /// </summary>
+    /// <returns>The validation result.</returns>
+    internal override ValidationResult Validate()
+    {
+        var result = base.Validate();
+
+        try
+        {
+            var deserializer = new DeserializerBuilder().Build();
```
dejobratic, line 22: Can we have the deserializer as a static field, so that it's not created each time we create a Yaml type?

tests/Nox.Types.Tests/EntityFrameworkTests/NoxTypesEntityFrameworkTests.cs
```diff
@@ -166,11 +191,17 @@ public void AddedItemShouldGetGeneratedId()
         item.CreateDate.Should().Be(DateTime.From(new System.DateTime(2023, 01, 01)));
         item.DateTimeDuration.Value.Should().Be(new TimeSpan(10, 5, 2, 1));
         item.Nuid.Value.Should().Be(NuidDefinition.NuidValue);
-        Assert.Equal(newItem.AverageTemperatureInCelsius.Value, item.AverageTemperatureInCelsius?.Value);
-        Assert.Equal(newItem.AverageTemperatureInCelsius.Unit, item.AverageTemperatureInCelsius?.Unit);
-        Assert.Equal(Sample_Uri, item.Uri.Value.AbsoluteUri);
         AssertStreetAddress(streetAddress, item.StreetAddress);
-        Assert.Equal(JsonSerializer.Serialize(streetAddress), item.StreetAddressJson.Value);
+        // Assert.Equal(newItem.AverageTemperatureInCelsius.Value, item.AverageTemperatureInCelsius?.Value);
```
dejobratic, line 195: Why are these assertions commented?

tests/Nox.Types.Tests/EntityFrameworkTests/Models/Country.cs
```diff
@@ -169,4 +169,9 @@ public sealed class Country
     /// Gets or sets the Flag.
     /// </summary>
     public Image Flag { get; set; } = null!;
+    
+    /// <summary>
+    /// Gets or sets the CitiesCounties.
+    /// </summary>
+    public Yaml CitiesCounties { get; set; } = null!;
```
dejobratic, line 176: Can we rename the property to CitiesCountiesYaml?

dejobratic (changes requested):
Looks good. Added some minor comments.

tests/Nox.Types.Tests/EntityFrameworkTests/Models/Country.cs
msadikkoseaupa: I would not prefer type as suffix in properties.

tests/Nox.Types.Tests/EntityFrameworkTests/NoxTypesEntityFrameworkTests.cs
msadikkoseaupa: Redundant lines were removed. They had been already rewritten with fluent assertions

src/Nox.Types/Types/Yaml/Yaml.cs
msadikkoseaupa: Requried changes were applied

src/Nox.Types.EntityFramework/Types/Yaml/YamlConverter.cs
```diff
@@ -0,0 +1,11 @@
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+
+namespace Nox.Types.EntityFramework.Types;
+
+/// <summary>
+/// Converts a <see cref="Yaml"/> to and from a <see cref="string"/> for Entity Framework.
+/// </summary>
+public class YamlConverter: ValueObjectConverter<Yaml, string>
```
rochar, line 8: we will remove the ValueObjectConverter since its creating an instance to do the FromDatabase

@gjorgji-grgovski-authoritypartners I'm approving but then we need to take car of this converter also

rochar (approved):
looks good, the converter needs to change in a following iterations

dejobratic (approved):
Good job.

PR #310, opened 2023-07-19
Implemented Phone Number type #68

dejobratic:
- Added PhoneNumber value object with string as underlying type.
- PhoneNumber format is validated against a RegEx.


src/Nox.Types/Types/PhoneNumber/PhoneNumber.cs
```diff
@@ -1,9 +1,28 @@
+using System.Text.RegularExpressions;
+using System;
+
 namespace Nox.Types;
 
+/// <summary>
+/// Represents a Nox <see cref="PhoneNumber"/> type and value object.
+/// </summary>
+public sealed class PhoneNumber : ValueObject<string, PhoneNumber>
```
rochar, line 9: SimpleTypes require a converter, and alo add a test in EF just to see it working

rochar (changes requested):
looks good, small comment

src/Nox.Types/Types/PhoneNumber/PhoneNumber.cs
dejobratic: Missing converter has been added.

rochar (approved):

PR #313, opened 2023-07-19
Remove the early exitter

fkucuk:
I believe this is a bug. When there is multiple user secret values, it was grabbing only the first one. Is it okay if we remove this line @jan-schutte ?


jan-schutte (commented):
@Fatih, thanx, was fixed in a PR yesterday

rochar:
Closing the PR per @jan-schutte  comment

PR #316, opened 2023-07-19
Get By Id simple type key

alexandervlasenko:


rochar (approved):
Looks good, in the future we need to update the generator to se liquid templates to see if its easier

ILoveMyTortoise (approved):

PR #318, opened 2023-07-19
MP-7468: User Nox Type - definition and implementation

originalethanhunt:


src/Nox.Types.EntityFramework/Types/User/UserConverter.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+using System.Linq.Expressions;
+
+namespace Nox.Types.EntityFramework.Types;
+
+/// <summary>
+/// The user converter.
+/// </summary>
+public class UserConverter : ValueObjectConverter<User, string>
```
dejobratic, line 9: Since ValueObjectConverter has been deleted we need to change the implementation of this converter to inherit from EFs ValueConverter.

src/Nox.Types.EntityFramework/Types/Email/EmailConverter.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Nox.Types.EntityFramework.Types;
+
+/// <summary>
+/// The email converter.
+/// </summary>
+public class EmailConverter : ValueObjectConverter<Email, string>
```
dejobratic, line 6: Since ValueObjectConverter has been deleted we need to change the implementation of this converter to inherit from EFs ValueConverter.

src/Nox.Types/Types/User/User.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        switch (_userTypeOptions.UserFormatType)
+        {
+            case UserFormatType.Guid:
+                if(!Guid.TryParse(Value, out var guidResult))
+                {
+                    result.Errors.Add(new ValidationFailure(nameof(Value), $"Could not create a Nox User type {Value} as it is not a valid Guid."));
+                }
+                break;
+            case UserFormatType.Email: Email.From(Value);break;
```
dejobratic, line 61: I believe the requirement was not to be referencing other Nox types. Instead, we should copy the relevant implementation.

tests/Nox.Types.Tests/Types/User/UserTests.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        userTypeOptions.MinLength = 1;
+        userTypeOptions.MaxLength = 511;
+        userTypeOptions.UserFormatType = UserFormatType.Guid;
+    }
+
+    [Fact]
+    public void User_Constructor_SpecifyingMaxLength_WithLongerLengthInput_ThrowsValidationException()
+    {
+        Assert.Throws<TypeValidationException>(() => _ =
```
dejobratic, line 42: Can we assert the exceptions using FluentAssertions?
We should assert that the exception has been thrown and that it contains expected errors.

dejobratic (commented):
Looks good. Added some comments.

src/Nox.Types/Types/User/User.cs
```diff
@@ -0,0 +1,83 @@
...
+/// The user nox type.
+/// </summary>
+public sealed class User : ValueObject<string, User>
+{
+    private UserTypeOptions _userTypeOptions = new();
+
+    /// <summary>
+    /// Gets or sets the value.
+    /// </summary>
+    public override string Value { protected set => base.Value = value.ToLowerInvariant(); }
```
rochar, line 18: not sure about the implications of this ToLowerInvariant

rochar (approved):
Looks good

src/Nox.Types/Types/User/User.cs
originalethanhunt: The user id needs to be case insensitive for comparison. I don't like storing things in upper case here. So to lower invariant primarily on English characters should work, unless we are allowing Unicode characters in the username. That's certainly something worth considering.

src/Nox.Types.EntityFramework/Types/User/UserConverter.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+using System.Linq.Expressions;
+
+namespace Nox.Types.EntityFramework.Types;
+
+/// <summary>
+/// The user converter.
+/// </summary>
+public class UserConverter : ValueConverter<User, string>
```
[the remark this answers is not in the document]
originalethanhunt: Indeed we lose it and that was a concern initially, but since value objects are immutable and they are validated at the time when they're stored, the type really doesn't matter but again is significant from the auditing context.

PR #329, opened 2023-07-20
Implemented Area type database configuration

dejobratic:
- Added AreaTypeOptions with MinValue, MaxValue and Unit to Area type;
- Added AreaDatabaseConfigurator that selects a converter based on the Unit in AreaTypeOptions;
- Updated tests for model generation;


schemas/domainCommand.json
rochar, line 1: Looks good, can you update our SampleWebApp nox solution and use area in this property:

```
  - name: AreaInSquareKilometres
          description: Country area in square kilometers
          type: number
          numberTypeOptions:
            minValue: 0
            maxValue: 20000000 # 20,000,000 > Russia
          isRequired: true
```

Update SeedData, and run the sample to see everything is ok (Including Odata Get Countries)

rochar (changes requested):
Looks goos, just one additional request

rochar (approved):

PR #330, opened 2023-07-20
Features/212 nuid construction

AndreyDegtyarev:
Introduced code generation for Nuid type.
Added validation for NuidTypeOptions to check whether entity properties exist.


rochar (changes requested):
please commit generated code, so I can continue the review

src/Nox.Generator/Common/TemplateScriptsBridges/NoxSolutionCodeGeneratorStateBridge.cs
```diff
@@ -14,6 +14,12 @@ public static void AddFunctions(TemplateContext context, NoxSolutionCodeGenerato
             scriptObject1.Import("CodeGeneratorStateGetForeignKeyPropertyName", new Func<NoxSimpleTypeDefinition,
                 string>((key) => codeGeneratorState.GetForeignKeyPropertyName(key.EntityTypeOptions!.Entity)));
 
+            scriptObject1.Import("CodeGeneratorPrivateFieldName", new Func<NoxSimpleTypeDefinition,
```
rochar, line 17: we will remove the option of include scriptObject, to make the template explicit and simple... this needs to be re think to be done with script objects

tests/Nox.Generator.Tests/ExpectedGeneratedFiles/Country.expected.g.cs
rochar, line 1: Nuid should be a get property, private to be set from the database without any logic 

Consider how this will be used by the application (Services / API), someone creates an Entity the controller or messaging receives this request, delegates to the Application Service that creates an Entity, how can it generate the Id?

And for the update use Case?

This should be immutable after the creation, update can/should only validate

rochar (changes requested):

tests/Nox.Tests/NuidTypeTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            };
+
+            entity.EnsureId();
+
+            _dbContext.TestEntityWithNuids.Add(entity);
+            _dbContext.SaveChanges();
+
+            var dbEntity = _dbContext.TestEntityWithNuids.First(x => x.Name == Text.From(nameValue));
+
+            Assert.Equal(entity.Id, dbEntity.Id);
```
rochar, line 33: Use Fluent Assertions

tests/Nox.Generator.Tests/ExpectedGeneratedFiles/Country.expected.g.cs
```diff
@@ -17,7 +17,23 @@ public partial class Country : AuditableEntityBase
...
+    		if(Id is null)
+    		{
+    			Id = Nuid.From("Country." + string.Join(".", Name.Value.ToString(),FormalName.Value.ToString()));
+    		}
+    		else
+    		{
+    			var currentNuid = Nuid.From("Country." + string.Join(".", Name.Value.ToString(),FormalName.Value.ToString()));
+    			if(Id != currentNuid)
+    			{
+    				throw new ApplicationException("Immutable nuid property Id value is different since it has been initialized");
```
rochar, line 33: Create a specific exception for this Nox***Exception

src/Nox.Types.Abstractions/TypeOptions/NuidTypeOptions.cs
```diff
@@ -4,6 +4,7 @@ namespace Nox.Types;
 
 public class NuidTypeOptions
 {
+    public string? Prefix { get; set; }
```
rochar, line 7: Add a comment, explaining the puposes, its optional, and and will be the prefix if this is not set

src/Nox.Types.Abstractions/TypeOptions/NuidTypeOptions.cs
rochar, line 1: You need to add this options to NoxSimpleTypeDefinition and regenerate the schemas

rochar (changes requested):
Looks good, just a few comments

src/Nox.Types.Abstractions/TypeOptions/NuidTypeOptions.cs
AndreyDegtyarev: Has already added.

src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -125,20 +125,19 @@ internal bool ApplyDefaults()
 
         if (Relationships is not null)
         {
-#pragma warning disable S3267 // Loops should be simplified with "LINQ" expressions
-            foreach (var relationship in Relationships)
+            var relationships = Relationships
```
rochar, line 128: why are you change this in this PR?

code cleanup needs to be done in its own PR

rochar (approved):

PR #333, opened 2023-07-20
Feature/mp 7730 country code3 db config

originalethanhunt:


src/Nox.Types.EntityFramework/Types/CountryCode3/CountryCode3DatabaseConfigurator.cs
```diff
@@ -0,0 +1,30 @@
...
+    public void ConfigureEntityProperty(
+        NoxSolutionCodeGeneratorState noxSolutionCodeGeneratorState,
+        EntityTypeBuilder builder,
+        NoxSimpleTypeDefinition property,
+        Entity entity,
+        bool isKey)
+    {
+        builder
+            .Property(property.Name)
+            .IsRequired(isKey || property.IsRequired)
```
gjorgji-grgovski-authoritypartners, line 22: We don't need to check "isKey" anymore, based on discussion with @rochar, so please use only "property.IsRequired"

gjorgji-grgovski-authoritypartners:
@originalethanhunt can we also change the Type to use a HashSet of all possible CountryCodes instead of using Regex?

gjorgji-grgovski-authoritypartners (approved):
Looks good with a small comment, @originalethanhunt please resolve it and mark it as Ready for Review

rochar (approved):
Good job, just a typo to fix

gjorgji-grgovski-authoritypartners (approved):

PR #340, opened 2023-07-21
Feature/etl loader

jan-schutte:


fkucuk (commented):
Can we use `[Fact(Skip="reason")]` attribute if the code is building?

src/DataConnections/DataConnections.csproj
rochar, line 1: Add Treat Warning as Errors, If we generate a packager  add also the Package configuration

rochar (changes requested):
I would like to discuss with you the new components and dependencies

PR #343, opened 2023-07-24
MP-7357 [BS-8103] NOX - Define and implement custom type "DateTimeSch…

becicajla:
…edule"

### DIFF
--- a/src/Nox.Types/Types/DateTimeSchedule/DateTimeSchedule.cs
+++ b/src/Nox.Types/Types/DateTimeSchedule/DateTimeSchedule.cs
@@ -1,0 +1,175 @@
+﻿using System.Globalization;
+using System;
+using System.Collections.Generic;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace Nox.Types;
+
+public class DateTimeSchedule : ValueObject<(System.DateTime StartDate, System.DateTime EndDate), DateTimeSchedule>
+{
+    private DateTimeScheduleTypeOptions _dateTimeScheduleTypeOptions = new();
+
+    /// <summary>
+    /// Gets the start of the date time range scheduled.
+    /// </summary>
+    public System.DateTime StartDate
+    {
+        get => Value.StartDate;
+        private set => Value = (StartDate: value, EndDate: Value.EndDate);
+    }
+
+    /// <summary>
+    /// Gets the end of the date time range scheduled.
+    /// </summary>
+    public System.DateTime EndDate
+    {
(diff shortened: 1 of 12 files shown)


src/Nox.Types/Types/DateTimeSchedule/DateTimeSchedule.cs
```diff
@@ -0,0 +1,175 @@
...
+
+    /// <summary>
+    /// Returns a string representation of the custom date and time using <see cref="CultureInfo.InvariantCulture"/>.
+    /// </summary>
+    public override string ToString()
+    {
+        return $"Start: {Value.StartDate}, End: {Value.EndDate}, Frequency: {_dateTimeScheduleTypeOptions.Frequency}, Frequency value: {_dateTimeScheduleTypeOptions.FrequencyValue}";
+    }
+
+    public DateTimeSchedule GetNextScheduledDate()
```
danield-ap, line 88: I think `GetNextScheduleDate` should just return the next Date, not a new Schedule object.

src/Nox.Types/Types/DateTimeSchedule/DateTimeSchedule.cs
```diff
@@ -0,0 +1,175 @@
+﻿using System.Globalization;
+using System;
+using System.Collections.Generic;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace Nox.Types;
+
+public class DateTimeSchedule : ValueObject<(System.DateTime StartDate, System.DateTime EndDate), DateTimeSchedule>
```
danield-ap, line 8: We also need to save the information about the frequency.

tests/Nox.Types.Tests/EntityFrameworkTests/Configuration/CountryDbConfiguration.cs
```diff
@@ -46,6 +46,7 @@ public void Configure(EntityTypeBuilder<Country> builder)
         builder.OwnsOne(e => e.LatLong).Ignore(p => p.Value);
         builder.OwnsOne(e => e.GrossDomesticProduct).Ignore(p => p.Value);
         builder.OwnsOne(e => e.DateTimeRange).Ignore(p => p.Value);
+        builder.OwnsOne(e => e.DateTimeSchedule).Ignore(p => p.Value);
```
danield-ap, line 49: I think we should have a converter, since we are using converter even for simpler types.
We should check with Ricardo which types won't need a Converter.

PR #347, opened 2023-07-24
Fixing sql server test base

emir-sehovic1:


ILoveMyTortoise (approved):

ILoveMyTortoise:
It should be fine for now as long as it works, we will have significant changes there soon anyway

PR #356, opened 2023-07-25
Refactor/loggin and monitoring

jan-schutte:
- improve logging and monitoring


src/Nox.Abstractions/Logging/LoggingOptionsBuilder.cs
rochar, line 1: I do prefer where on Nox.Abstraction

rochar (changes requested):

PR #357, opened 2023-07-25
Database configuration for "CultureCode"

msadikkoseaupa:


rochar (approved):
Looks good, you also need to change the converter to use CultureCode.FromDatabase to skip validation

PR #360, opened 2023-07-25
MP-7821   [BS-8103] NOX - Database configuration for "Email"

becicajla:


src/Nox.Types.EntityFramework/Types/Email/EmailDatabaseConfigurator.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Nox.Types.EntityFramework.Types;
+public class EmailDatabaseConfigurator : INoxTypeDatabaseConfigurator
+{
+    public NoxType ForNoxType => NoxType.Email;
+
+    public bool IsDefault => true;
+
+    public void ConfigureEntityProperty(NoxSolutionCodeGeneratorState noxSolutionCodeGeneratorState, EntityTypeBuilder builder, NoxSimpleTypeDefinition property, Entity entity, bool isKey)
+    {
+        builder
```
gjorgji-grgovski-authoritypartners, line 16: What would be the column type for SqlServer and Postgres? Do we maybe need to set some additional properties like IsUnicode?

gjorgji-grgovski-authoritypartners: Do we need to set the length? What is the default length in this case?
Maybe test this with the SampleWebApp and generate Migrations

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types.EntityFramework/Types/Email/EmailDatabaseConfigurator.cs
becicajla: @gjorgji-grgovski-authoritypartners the default value in sql for this column is nvarchar(max). I left it as unicode, but restricted column to 255 characters

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):

PR #361, opened 2023-07-25
WIP: add interim one to one change

ILoveMyTortoise:


src/Nox.Types.EntityFramework/Abstractions/NoxDatabaseConfigurator.cs
```diff
@@ -57,6 +57,16 @@ public abstract class NoxDatabaseConfigurator : INoxDatabaseConfigurator
         {
             foreach (var relationshipToCreate in relationshipsToCreate)
             {
+                // ManyToMany does not need to be handled
+                // Handle ZeroOrOne or ExactlyOne scenario with foreign key.
+                if (relationshipToCreate.ShouldBeMapped)
```
rochar, line 62: I think we could be explicitly and also check that  the relation is ZeroOrOne or ExactlyOne

Otherwise will be easy to introduce a issue

rochar (approved):

rochar (approved):

PR #372, opened 2023-07-26
266 Create NoxType VatNumber

anna-naumova25:
VatNumber Nox type created with using Regex validation. Can be updated to Nox.Reference if needed.

### DIFF
--- a/src/Nox.Types.Abstractions/TypeOptions/VatNumberOptions.cs
+++ b/src/Nox.Types.Abstractions/TypeOptions/VatNumberOptions.cs
@@ -2,5 +2,7 @@
 
 public class VatNumberOptions : INoxTypeOptions
 {
-    public string DefaultCountryCode { get; internal set; } = "GB";
+
+    private static readonly string DefaultCountryCode = "GB";
+    public string CountryCode { get; set; } = DefaultCountryCode;
 }
--- a/src/Nox.Types.Abstractions/TypeOptions/VatNumberOptions.cs
+++ b/src/Nox.Types.Abstractions/TypeOptions/VatNumberOptions.cs
@@ -1,0 +1,6 @@
+﻿namespace Nox.Types;
+
+public class VatNumberOptions : INoxTypeOptions
+{
+    public string DefaultCountryCode { get; internal set; } = "GB";
+}
--- a/src/Nox.Types/Types/VatNumber/VatNumber.cs
+++ b/src/Nox.Types/Types/VatNumber/VatNumber.cs
@@ -1,9 +1,125 @@
+using System;
+using System.Text.RegularExpressions;
+
 namespace Nox.Types;
 
-    /// <summary>
-    /// Represents a Nox <see cref="VatNumber"/> type and value object.
-    /// </summary>
-    /// <remarks>Placeholder, needs to be implemented</remarks>
-    public sealed class VatNumber : ValueObject<(string VatNumber, CountryCode2 countryCode2), VatNumber>
+/// <summary>
+/// Represents a Nox <see cref="VatNumber"/> type and value object.
+/// </summary>
+/// <remarks>Placeholder, needs to be implemented</remarks>
+public sealed class VatNumber : ValueObject<(string VatNumberValue, CountryCode2 CountryCode), VatNumber>
+{
+
+    #region VatNumber Regex
+
+    private static readonly Regex _vatNumberRegex = new(
+        @"^(
+(AT)?U[0-9]{8} |                              # Austria
+(BE)?0[0-9]{9} |                              # Belgium
+(BG)?[0-9]{9,10} |                            # Bulgaria
+(CY)?[0-9]{8}L |                              # Cyprus
+(CZ)?[0-9]{8,10} |                            # Czech Republic
+(DE)?[0-9]{9} |                               # Germany
+(DK)?[0-9]{8} |                               # Denmark
+(EE)?[0-9]{9} |                               # Estonia
+(EL|GR)?[0-9]{9} |                            # Greece
+(ES)?[0-9A-Z][0-9]{7}[0-9A-Z] |               # Spain
+(FI)?[0-9]{8} |                               # Finland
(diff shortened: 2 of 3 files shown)


src/Nox.Types.Abstractions/TypeOptions/VatNumberOptions.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Nox.Types;
+
+public class VatNumberOptions : INoxTypeOptions
+{
+    public string DefaultCountryCode { get; internal set; } = "GB";
```
rochar, line 5: we are following a convention like this:

```
    public static readonly string DefaultCountryCode = "GB";
    public string CountryCode{ get; set; } = DefaultCountryCode ;
```

src/Nox.Types/Types/VatNumber/VatNumber.cs
```diff
@@ -1,9 +1,125 @@
...
+    /// <summary>
+    /// Creates a new instance of the <see cref="VatNumber"/> class with the specified values.
+    /// </summary>
+    /// <param name="value">The VAT number value.</param>
+    /// <param name="countryCode">The two-letter country codes (ISO alpha-2).</param>
+    /// <returns>A new instance of the <see cref="VatNumber"/> class.</returns>
+    public static VatNumber From(string value, CountryCode2 countryCode) => 
+        From((value, countryCode));
+
+    public static VatNumber From(string value, VatNumberOptions typeOptions) =>
```
rochar, line 78: for the From Overloads, I would suggest

```
From(string vatNumber, string countryCode2) => From(vatNumber , new VatNumberOptions(){CountryCode2 = countryCode2})
From(string vatNumber, VatNumberOptions typeOptions)

```
Do not combine VatNumberOptions  and CountryCode, because they may contradict, and that is not the intention

src/Nox.Types/Types/VatNumber/VatNumber.cs
```diff
@@ -1,9 +1,125 @@
...
+    /// <summary>
+    /// Validates a <see cref="VatNumber"/> object.
+    /// </summary>
+    /// <returns>A validation result indicating whether the <see cref="VatNumber"/> value is valid or not.</returns>
+    internal override ValidationResult Validate()
+    {
+        var result = base.Validate();
+
+        var vatNumberStr = VatNumberString();
+        if (!_vatNumberRegex.IsMatch(vatNumberStr))
```
rochar, line 108: And if the Country does not have a regular expression what will happen?

src/Nox.Types/Types/VatNumber/VatNumber.cs
```diff
@@ -1,9 +1,125 @@
...
+/// <summary>
+/// Represents a Nox <see cref="VatNumber"/> type and value object.
+/// </summary>
+/// <remarks>Placeholder, needs to be implemented</remarks>
+public sealed class VatNumber : ValueObject<(string VatNumberValue, CountryCode2 CountryCode), VatNumber>
+{
+
+    #region VatNumber Regex
+
+    private static readonly Regex _vatNumberRegex = new(
```
rochar, line 15: Would not be more simple to have a dictionary, key country code and value regex?

tests/Nox.Types.Tests/Types/VatNumberTests/VatNumberTests.cs
```diff
@@ -4,8 +4,107 @@ namespace Nox.Types.Tests.Types;
...
+    {
+        //Arrange
+        var vatNumber1 = VatNumber.From("44403198682", CountryCode2.From("FR"));
+        var vatNumber2 = VatNumber.From("44403198682", CountryCode2.From("FR"));
+
+        //Act
+        var result = vatNumber1.Equals(vatNumber2);
+
+        //Assert
+        Assert.True(result);
```
rochar, line 33: See and use Fluent.Assertions, we are using for assertions

rochar (changes requested):
looks good a few comment's

src/Nox.Types.Abstractions/TypeOptions/VatNumberOptions.cs
anna-naumova25: Have updated the code to follow the above mentioned convention

src/Nox.Types/Types/VatNumber/VatNumber.cs
anna-naumova25: I've updated the overloads to the suggested ones. But I have a small question about it - don't we expect that the user will use DefaultCountryCode and then insert VatNumber with other CountryCode?

src/Nox.Types/Types/VatNumber/VatNumber.cs
anna-naumova25: Have updated to Dictionary

tests/Nox.Types.Tests/Types/VatNumberTests/VatNumberTests.cs
anna-naumova25: Have updated tests to using FluentAssertions. One short question - is it a general rule for all the Nox libs to use just FluentAssertions for tests? I'm asking because I saw both Assert.True and FluentAssertions.

rochar: In General,  new stuff with do it in Fluent, other parts we migrate whenever its possible

src/Nox.Types/Types/VatNumber/VatNumber.cs
rochar: If I understand the question, the user needs to add a VatNumber and express in what countryCode, via Options or explicit parameter, the default is just a facility to not express the country code and use the default in the options. We could also create a overload From (VatNumber), that would use the default from options.

rochar (approved):

src/Nox.Types/Types/VatNumber/VatNumber.cs
rochar: we need to discuss this, I will approve the PR and merge

PR #388, opened 2023-07-27
MP-7849: Ipaddress dbconfig

originalethanhunt:


src/Nox.Types.EntityFramework/Types/IpAddress/IpAddressDatabaseConfigurator.cs
```diff
@@ -0,0 +1,29 @@
...
+public class IpAddressDatabaseConfigurator : INoxTypeDatabaseConfigurator
+{
+    /// <inheritdoc/>
+    public NoxType ForNoxType => NoxType.IpAddress;
+
+    /// <inheritdoc/>
+    public bool IsDefault => true;
+
+    /// <inheritdoc/>
+    public void ConfigureEntityProperty(NoxSolutionCodeGeneratorState noxSolutionCodeGeneratorState, EntityTypeBuilder builder, NoxSimpleTypeDefinition property, Entity entity, bool isKey)
```
gjorgji-grgovski-authoritypartners, line 19: https://dirask.com/static/bucket/1659620131453-r1AjPkeY5a--image.png

As the max length of an IPv6 Address is 45 characters we can perhaps set a fixed length to 63

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types.EntityFramework/Types/IpAddress/IpAddressDatabaseConfigurator.cs
gjorgji-grgovski-authoritypartners: @rochar can you please share your thought about the length, as I would prefer to always limit the length if we know the max length?

rochar: agree, why 63 and not 45? or 46 https://www.ibm.com/docs/en/psww2500/2.3.3?topic=linux-ip-address-attributes... we can also add this validation to the type, its easy and would fail with a proper message?

gjorgji-grgovski-authoritypartners: image

because of this message from Andre, I would say so.

rochar: in my understanding this is for holding bits 2^6 => 64 bits, 2^7 => 128 bits , we are setting up chars (1 byte size)...

rochar (approved):

dejobratic (approved):
Looks good.

PR #389, opened 2023-07-27
MP-7793 [BS-8103] NOX - Database configuration for "Date"

emir-sehovic1:
- Implemented db configuration for Date nox type
- Upgraded underlying type to DateOnly since Nox.Types was upgraded to net7.0
- Added postgres integration tests


gjorgji-grgovski-authoritypartners (approved):

src/Nox.EntityFramework.Postgres/PostgresDateDatabaseConfigurator.cs
rochar, line 1: I do not think we need any specific database configuration, EF should take car of all cases, for SQL for example: https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings

since the converter already states is datetime

rochar (changes requested):
looks good, one comment to validate

rochar (approved):

PR #392, opened 2023-07-28
Features/387 global error handling sample web app

AndreyDegtyarev:
[OData] Ensure that Odata runtime exceptions are not swallowed


src/SampleWebApp/Extensions/ApplicationExtensions.cs
rochar, line 1: we can add this by default in ServiceCollectionExtension.AddNoxLib

rochar (changes requested):
Looks good, all can be a Nox implemenation, the sample webApp use this by default

for now we want to make Nox adoption and feature to be used only with AddNox() option

src/SampleWebApp/Extensions/ApplicationExtensions.cs
AndreyDegtyarev: Moved to UseNox() extension method since middleware should be added in application's pipeline

rochar (approved):

PR #399, opened 2023-07-29
 MP-7877 latlong db config

msadikkoseaupa:


gjorgji-grgovski-authoritypartners:
The title of the PRs should start with "MP-XXXX ........."

src/Nox.Types.EntityFramework/Types/LatLong/LatLongConfigurator.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <param name="entity">The Entity to which the property belongs.</param>
+    /// <param name="isKey">A flag indicating whether the property is a key property.</param>
+    public void ConfigureEntityProperty(
+        NoxSolutionCodeGeneratorState noxSolutionCodeGeneratorState,
+        EntityTypeBuilder builder,
+        NoxSimpleTypeDefinition property,
+        Entity entity,
+        bool isKey)
+    {
+        builder
```
gjorgji-grgovski-authoritypartners, line 41: We would need to set the type of the column and it's Precision so that we make sure we can keep up to 8/9 decimal places

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types.EntityFramework/Types/LatLong/LatLongConfigurator.cs
```diff
@@ -0,0 +1,54 @@
...
+using Nox.Solution;
+using Nox.Types.EntityFramework.Abstractions;
+
+namespace Nox.Types.EntityFramework.Types;
+
+/// <summary>
+/// Configurator for the <see cref="LatLong"/> Nox type in the database.
+/// This class implements the INoxTypeDatabaseConfigurator interface to provide configuration details for the  <see cref="LatLong"/> Nox type in the database.
+/// </summary>
+public class LatLongConfigurator : INoxTypeDatabaseConfigurator
```
dejobratic, line 12: Can we rename this class to LatLongDatabaseConfigurator, to be consistent with all types?

src/Nox.Types.EntityFramework/Types/LatLong/LatLongConfigurator.cs
```diff
@@ -0,0 +1,54 @@
...
+        NoxSolutionCodeGeneratorState noxSolutionCodeGeneratorState,
+        EntityTypeBuilder builder,
+        NoxSimpleTypeDefinition property,
+        Entity entity,
+        bool isKey)
+    {
+        builder
+            .OwnsOne(typeof(LatLong), property.Name)
+            .Ignore(nameof(LatLong.Value))
+            .Property(nameof(LatLong.Latitude));
```
dejobratic, line 44: We need to confirm whether we want to enforce certain decimal precision on Latitude and Longitude properties. If not, the last statement can be removed.

dejobratic (changes requested):
Just a few comments.

src/Nox.Types.EntityFramework/Types/LatLong/LatLongConfigurator.cs
msadikkoseaupa: Required change was applied

src/Nox.Types.EntityFramework/Types/LatLong/LatLongConfigurator.cs
dejobratic: Yes, we would need to set up the precision here.
According to https://learn.microsoft.com/en-us/sql/t-sql/data-types/float-and-real-transact-sql?view=sql-server-ver16 for SqlServer we would just need the 7 digit precision. We also need to check for other database providers.

src/Nox.Types.EntityFramework/Types/LatLong/LatLongConfigurator.cs
msadikkoseaupa: Required changes were applied

dejobratic: Please also make sure that the integration tests for each database provider are passing and that we have the required precision in database.

src/Nox.Types.EntityFramework/Types/LatLong/LatLongConfigurator.cs
dejobratic: Please also make sure that the integration tests for each database provider are passing and that we have the required precision in database.

dejobratic (approved):

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #400, opened 2023-07-31
MP-7913 - Percentage Db configuration

denisvieira-dev:


src/Nox.Types.EntityFramework/Types/Percentage/PercentageDatabaseConfigurator.cs
```diff
@@ -0,0 +1,29 @@
...
+        Entity entity,
+        bool isKey)
+    {
+        var typeOptions = property.PercentageTypeOptions ?? new PercentageTypeOptions();
+
+        builder
+            .Property(property.Name)
+            .IsRequired(property.IsRequired)
+            .HasMaxLength(typeOptions.Digits)
+            .HasConversion<PercentageConverter>();
```
dejobratic, line 25: Can we update the PercentageConverter to use the Percentage.FromDatabase method when converting db value to value object?

dejobratic (changes requested):
Looks good. Just a small comment.

src/Nox.Types.EntityFramework/Types/Percentage/PercentageDatabaseConfigurator.cs
danield-ap: PercentageConverter is using `FromDatabase`.

rochar (approved):

dejobratic (approved):

PR #416, opened 2023-08-01
353 Generate Public Property for Formula Attribute Type

anna-naumova25:


src/Nox.Types.EntityFramework/Types/Formula/FormulaDatabaseConfigurator.cs
```diff
@@ -0,0 +1,28 @@
...
+    public virtual bool IsDefault => true;
+
+    public void ConfigureEntityProperty(
+        NoxSolutionCodeGeneratorState noxSolutionCodeGeneratorState,
+        EntityTypeBuilder builder,
+        NoxSimpleTypeDefinition property,
+        Entity entity,
+        bool isKey)
+    {
+        builder.Ignore(nameof(Formula));
```
rochar, line 20: should ignore the name of the property, not the type? property.Name?

rochar (changes requested):
looks good, small comment to review

src/Nox.Types.EntityFramework/Types/Formula/FormulaDatabaseConfigurator.cs
anna-naumova25: Updated

PR #418, opened 2023-08-01
Feature/#311 fix one to one

ILoveMyTortoise:


src/Nox.Solution/Models/Domain/Entities/Relationships/EntityRelationship.cs
```diff
@@ -38,6 +38,10 @@ public class EntityRelationship : DefinitionBase
     [Description("This boolean controls whether this side of the relationship is exposed in the generated code and ODATA endpoints.")]
     public bool CanNavigate { get; internal set; } = true;
 
+    [Title("Determines whether this side of the relationship should contain contain a foreign key.")]
```
rochar, line 41: this is to be ignored, this is not a user option but a computed property, use YamlIgnore and remove the other attribute

src/Nox.Solution/NoxSolution.cs
```diff
@@ -29,45 +29,47 @@ internal void Validate()
...
+                RelationshipEntityType = getTypeByNameFunc(relationship.Entity)!
+            });
+        }
+
+        return fullRelationshipModels;
+    }
 
-            var pairRelationship = relationship.Related.EntityRelationship;
-            if (pairRelationship != null)
+    public static bool ShouldGenerateForeignOnThisSide(EntityRelationship relationship)
```
rochar, line 42: mark has internal

rochar (approved):
looks good, but see my two comments and integrate the feedback

PR #423, opened 2023-08-02
MP-7863 - Database configuration for "JwtToken"

dejobratic:


danield-ap (approved):
Looking good.
I was wondering if we should have min/max size for the column, but it seems a jwt token can vary a lot in size.

rochar (approved):

PR #424, opened 2023-08-02
MP-7814 - Database configuration for "Distance"

dejobratic:
- Removed Measurement inheritance for Distance type;
- Updated conversion factors;
- Added database configurators;
- Updated unit tests;


src/Nox.Types/Types/Distance/Distance.cs
```diff
@@ -1,74 +1,175 @@
...
 using System.Collections.Generic;
-
+using System.Globalization;
 namespace Nox.Types;
 
-public class Distance : Measurement<Distance, DistanceUnit>
+/// <summary>
+/// Represents a Nox <see cref="Distance"/> type and value object.
+/// </summary>
+public class Distance : ValueObject<QuantityValue, Distance>
```
danield-ap, line 10: Why did this change since `Measurement` already implements a lot of the methods you have here?

src/Nox.Types/Types/Distance/Distance.cs
```diff
@@ -1,74 +1,175 @@
...
+    }
 
     protected override IEnumerable<KeyValuePair<string, object>> GetEqualityComponents()
     {
         yield return new KeyValuePair<string, object>(nameof(Value), ToKilometers());
     }
 
-    private static double CalculateDistanceInKilometers(LatLong origin, LatLong destination)
-        => new HaversineDistanceCalculator().Calculate(origin, destination);
+    public static Distance FromDatabase(QuantityValue distanceValue, DistanceTypeUnit distanceUnit)
```
danield-ap, line 143: Missing xml doc comment.

src/Nox.Types/Types/Distance/Distance.cs
dejobratic: There were a few reasons and one of them was that it was difficult to add type options and enum as unit for the type options with Measurement as base class. Therefore we decided to remove the inheritance.

src/Nox.Types/Types/Distance/Distance.cs
dejobratic: Comment was added. Thanks.

src/Nox.Types/Nox.Types.csproj
```diff
@@ -24,6 +24,7 @@
   <ItemGroup>
     <None Include="..\..\images\nox.png" Pack="true" PackagePath="" />
     <InternalsVisibleTo Include="Nox.Types.Tests" />
+    <InternalsVisibleTo Include="Nox.EntityFramework.SqlServer" />
```
rochar, line 27: @dejobratic this should not be needed, prefer to make dependencies public

rochar (approved):
looks good, please review my comment

src/Nox.Types/Nox.Types.csproj
dejobratic: Changed. Thanks.

danield-ap (approved):

PR #426, opened 2023-08-02
MP-7962 uri database config

msadikkoseaupa:
There is no standart for the uri length, different systems has different max length support. 
Should we add limit for the uri?


src/Nox.Types.EntityFramework/Types/Uri/UriConverter.cs
```diff
@@ -3,5 +3,5 @@
 namespace Nox.Types.EntityFramework.Types;
 public class UriConverter : ValueConverter<Uri, string>
```
dejobratic, line 4: Since the underlying type is System.Uri, should we use it here as well?

dejobratic (commented):
Looks good. Just a question related to converter.

src/Nox.Types.EntityFramework/Types/Uri/UriConverter.cs
dejobratic: We should use string, as it is in this example here.

dejobratic (approved):

rochar (approved):

PR #428, opened 2023-08-02
417 noxgenerator exclude soft delete entities for queries

anna-naumova25:


anna-naumova25:
@rochar I've just thought about adding IsDeleted property to IDeleteEntity. Maybe it would be a cleaner solution. Please let me know your thoughts.

src/Nox.Generator/Application/Queries/ByIdQueryGenerator.cs
```diff
@@ -24,6 +24,7 @@ public static void Generate(SourceProductionContext context, NoxSolutionCodeGene
                 .WithClassName($"Get{entity.Name}ByIdQuery")
                 .WithFileNamePrefix($"Queries")
                 .WithObject("entity", entity)
+                .WithObject("isVersioned", (entity.Persistence?.IsVersioned ?? true))
```
rochar, line 27: I know we have this on other generator, we would prefer not to have .WithObject and make it clear and obvious on the template {{if entity.Persistence.IsVersioned}}

src/Nox.Generator/Application/Queries/Query.template.cs
```diff
@@ -21,6 +21,7 @@ public class Get{{entity.PluralName}}QueryHandler : IRequestHandler<Get{{entity.
 
     public Task<IQueryable<O{{entity.Name}}>> Handle(Get{{entity.PluralName}}Query request, CancellationToken cancellationToken)
     {
-        return Task.FromResult((IQueryable<O{{entity.Name}}>)DataDbContext.{{entity.PluralName}});
+        var item = (IQueryable<O{{entity.Name}}>)DataDbContext.{{entity.PluralName}}{{if isVersioned}}.Where(r => !(r.Deleted == true)){{end}};
```
rochar, line 24: do we need r.Deleted == true? r.Deleted should be enough?

src/Nox.Generator/Application/Queries/QueryGenerator.cs
```diff
@@ -25,6 +25,7 @@ public static void Generate(SourceProductionContext context, NoxSolutionCodeGene
                 .WithClassName($"Get{entity.PluralName}Queries")
                 .WithFileNamePrefix($"Queries")
                 .WithObject("entity", entity)
+                .WithObject("isVersioned", (entity.Persistence?.IsVersioned ?? true))
```
rochar, line 28: same has previous comment

rochar (changes requested):
looks good, just two small comments to review

src/Nox.Generator/Application/Queries/Query.template.cs
anna-naumova25: r.Deleted won't work as Deleted is a nullable field and it cannot automatically convert bool? to bool.

src/Nox.Generator/Application/Queries/ByIdQueryGenerator.cs
anna-naumova25: Updated

rochar (approved):

rochar (approved):

PR #431, opened 2023-08-03
MP-7948 [BS-8103] NOX - Database configuration for "Weight"

emir-sehovic1:


src/Nox.Types.Abstractions/TypeDefinitions/NoxSimpleTypeDefinition.cs
```diff
@@ -98,6 +98,9 @@ public class NoxSimpleTypeDefinition
     [IfEquals("Type", NoxType.Length)]
     public LengthTypeOptions? LengthTypeOptions { get; set; }
 
+    [IfEquals("Type", NoxType.Weight)]
+    public WeightTypeOptions? WeightTypeOptions { get; set; }
```
dejobratic, line 102: Please regenerate the schema.

src/Nox.Types/Types/Weight/Weight.cs
```diff
@@ -1,51 +1,141 @@
...
 
 namespace Nox.Types;
 
 /// <summary>
 /// Represents a Nox <see cref="Weight"/> type and value object.
 /// </summary>
-public class Weight : Measurement<Weight, WeightUnit>
+public class Weight : ValueObject<QuantityValue, Weight>
 {
+    public const int QuantityValueDecimalPrecision = 6;
```
dejobratic, line 14: Since the internals are visible to SqlServer project, can we mark this field as internal?

dejobratic (changes requested):
Looks good. Just a few minor comments.

emir-sehovic1:
> Looks good. Just a few minor comments.

Thanks for feedback, I've updated the implementation

dejobratic (approved):

src/Nox.Types/Nox.Types.csproj
```diff
@@ -24,6 +24,7 @@
   <ItemGroup>
     <None Include="..\..\images\nox.png" Pack="true" PackagePath="" />
     <InternalsVisibleTo Include="Nox.Types.Tests" />
+	  <InternalsVisibleTo Include="Nox.EntityFramework.SqlServer" />
```
rochar, line 27: this should not be needed?

rochar (changes requested):
Looks good, one thing to review

src/Nox.Types/Nox.Types.csproj
emir-sehovic1: Property Weight.QuantityValueDecimalPrecision, and classes WeightConversion, WeightUnit were marked as internal, hence that was needed. But I've marked them now as public and removed this line as we discussed.

rochar (approved):

Issue #408, opened 2023-07-31
[Nox.Generators] [Odata] Refactor ODataModelGenerator extract EntitiyDto model to its own Generator


rochar:
ODataModelGenerator has many responsibilities and does not use templates to generate code. In the scope of this issue lets extract the {EntityName}Dto Model to its own generator and class, using templates

see EntityGenerator as an example

PR #435, opened 2023-08-04
Issues/408 enitity dto generator refactoring

AndreyDegtyarev:
* [Nox.Generators] [Odata] Refactor ODataModelGenerator extract OEntitiy model to its own Generator #407
[Nox.Generators] [Odata] Refactor ODataModelGenerator extract EntitiyDto model to its own Generator #408

Changed code generation. Moved  code generation logic to template


rochar (approved):
looks good, 

- we need to move the Generator to the Application (this is not only for presentation, this is for providing data from the Domain)

-  Update the name space to be ....Aplication.Dto
- Update the generator Name to be DtoGenerator

PR #438, opened 2023-08-04
MP-8192 - Enable Dependabot to automate updates on stable dependencies

emir-sehovic1:
- Added dependabot
- Scheduled to check for minor and patch nuget package updates daily at 07:00 AM UTC
- Updates are checked for all projects inside Nox.Generator solution (Nox.Types.Benchmarks, CoreInterfaces.csproj and Nox.PackageManager.csproj are not part of the solution at this moment hence they are not being checked)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "nuget"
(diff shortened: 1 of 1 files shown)


dejobratic (approved):
Looks good.

rochar (approved):

PR #452, opened 2023-08-07
Feature/408 refactoring nox generator

AndreyDegtyarev:
[Nox.Generators] [Odata] Refactor ODataModelGenerator extract OEntitiy model to its own Generator https://github.com/NoxOrg/Nox.Generator/issues/407
[Nox.Generators] [Odata] Refactor ODataModelGenerator extract EntitiyDto model to its own Generator https://github.com/NoxOrg/Nox.Generator/issues/408
Refactoring CodeGenerator.


rochar (approved):

src/Nox.Generator/Nox.Generator.cs
```diff
@@ -135,6 +114,41 @@ private void GenerateSource(SourceProductionContext context, ImmutableArray<(str
         _debug.GenerateSourceCode();
     }
 
+    private static Dictionary<NoxGeneratorKind, INoxCodeGenerator[]> CreateGeneratorInstances()
```
andresharpe, line 117: I propose we just scan the executing assembly for types of  INoxCodeGenerator and create them using reflection.

That way this code doesn't need to be maintained by hand and will reduce merge conflicts.

andresharpe (changes requested):

src/Nox.Generator/Nox.Generator.cs
rochar: I did ask him to do change this, to avoid reflection and Activator.CreateInstance on Build time. So its a trade-off we need to select what we want more!

PR #462, opened 2023-08-08
Feature/solution improvements

jan-schutte:


rochar (approved):
looks good, juts a comment

.nox/design/sample.solution.nox.yaml
```diff
@@ -10,7 +10,7 @@ description: Sample Nox solution yaml configuration
 
 variables:
    DATABASE_PROVIDER: slqServer   
-   DATABASE_SERVER: localhost
+   DATABASE_SERVER: SamplewebAppDb
```
alexandervlasenko, line 13: Are we sure that we want to have SamplewebAppDb as a default setup?

alexandervlasenko (approved):

rochar:
lets close it until we define what we want to do

PR #464, opened 2023-08-08
Feature/localization type changes

jan-schutte:


src/Nox.Solution/Models/Application/Localization/Localization.cs
```diff
@@ -0,0 +1,24 @@
...
+[Title("The localization settings for the solution.")]
+[Description("The supported localization and defaults for the solution.")]
+[AdditionalProperties(false)]
+public class Localization
+{
+    [Title("The list of culture codes supported by the solution.")]
+    [Description("The list of culture codes that the solution supports.")]
+    [Pattern(@"^[a-z]{2}-[A-Z]{2}$")]
+    [Required]
+    public IReadOnlyList<string> SupportedCultures { get; set; } = new List<string>() { "en-US" };
```
rochar, line 16: could we use CultureCode type to get valitaion (5 letters)

jan-schutte: Good idea, but it would create a dependency on Nox.Types.

PR #466, opened 2023-08-08
Feature/227 noxgenerators odata get with multiple primary keys

anna-naumova25:
Added multipleIdsType entity to SampleWebApp to illustrate how Get by multiple keys work. Initially it should be done in scope of AllNoxTypes, but while I worked on the feature, AllNoxTypes moved to databaseNumber key which cannot be used twice per model. So if you want me to move/change using of compound primary keys, please let me know.


src/Nox.Generator/Application/Dto/EntityDto.template.cs
```diff
@@ -17,8 +21,10 @@ namespace {{codeGeneratorState.ApplicationNameSpace}}.Dto;
 /// {{entity.Description}}.
 /// </summary>
 [AutoMap(typeof({{entity.Name}}CreateDto))]
+[PrimaryKey({{primaryKeyParams}})]
```
rochar, line 24: this change should not be necessary, the model and the primary keys are being set by the Databaseconfiguration

I would expect changes only on the controller level

rochar (changes requested):
looks good, one comment to discuss

rochar:
can we also create a test on Nox.Generators asserting the result of a controller with multiple keys

src/Nox.Generator/Application/Dto/EntityDto.template.cs
anna-naumova25: Unfortunately, the following exception is thrown on any API call if this line of code is removed:

{
  "statusCode": 500,
  "Message": "The entity type 'MultipleIdsTypeDto' has multiple properties with the [Key] attribute. Composite primary keys configured by placing the [PrimaryKey] attribute on the entity type class, or by using 'HasKey' in 'OnModelCreating'."
} 

Even considering that we have HasKey, it still doesn't work without PrimaryKey

anna-naumova25: As discussed, updated to HasKey

anna-naumova25:
Added controller test to Nox.Generator.

Updated AllNoxTypes to using compound keys.

Will merge with main once I fix local build.

rochar:
@anna-naumova25  I see the deleted migration files, I do no see the new ones?

when @alexandervlasenko approves it, please merge and delete the branch

alexandervlasenko (approved):
Please, push the updated db migrations. Otherwise everything is fine.

src/Nox.Generator/Application/Queries/ByIdQuery.template.cs
```diff
@@ -24,7 +24,16 @@ public class Get{{entity.Name}}ByIdQueryHandler: IRequestHandler<Get{{entity.Nam
     {    
         var item = DataDbContext.{{entity.PluralName}}
             .AsNoTracking()
-            .SingleOrDefault(r => {{if (entity.Persistence?.IsVersioned ?? true)}}!(r.Deleted == true) && {{end}}r.{{entity.Keys[0].Name}}.Equals(request.key));            
+            .SingleOrDefault(r =>                  
+            {{- for key in entity.Keys }}
+                r.{{key.Name}}.Equals(request.key{{key.Name}}) && 
+            {{- end -}}
+            {{- if (entity.Persistence?.IsVersioned ?? true)}}
+                !(r.Deleted == true)
```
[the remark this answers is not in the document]
rochar: since Delete is nullable that would create a wrong query, ignoring null values

PR #487, opened 2023-08-09
Extend test coverage for relationship

ILoveMyTortoise:


tests/Nox.Tests/DatabaseIntegrationTests/Models/Entities.TestEntityOneOrManyToExactlyOne.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    ///  (Required).
+    /// </summary>
+    public Nox.Types.Text TextTestField2 { get; set; } = null!;
+
+    /// <summary>
+    /// TestEntityOneOrManyToExactlyOne Test entity relationship to TestEntityExactlyOneToOneOrMany OneOrMany TestEntityExactlyOneToOneOrManies
+    /// </summary>
+    public virtual List<TestEntityExactlyOneToOneOrMany> TestEntityExactlyOneToOneOrManies { get; set; } = new();
```
AndreyDegtyarev, line 30: Can we make testable relational entities generic to cut down on property definition.? The same is actual for "single" naviagation property

ILoveMyTortoise: no, because we generate entities fully from scratch the same way we do in production scenario

alexandervlasenko (approved):

.nox/design/sample.solution.nox.yaml
```diff
@@ -10,10 +10,10 @@ description: Sample Nox solution yaml configuration
 
 variables:
    DATABASE_PROVIDER: slqServer   
-   DATABASE_SERVER: SamplewebAppDb
+   DATABASE_SERVER: 127.0.0.1
```
rochar, line 13: you need to revert this line, we need to double check with @jan-schutte 
its related to the new docker project

src/Nox.Solution/Models/Domain/Entities/Relationships/EntityRelationship.cs
```diff
@@ -41,6 +41,15 @@ public class EntityRelationship : DefinitionBase
     [YamlIgnore]
     public bool ShouldGenerateForeignOnThisSide => NoxSolution.ShouldGenerateForeignOnThisSide(this);
 
+    [YamlIgnore]
+    public bool HasRelationshipWithSingularEntity => NoxSolution.HasRelationshipWithSingularEntity(this);
```
rochar, line 45: this circular dependency should be avoid, all this 3 methods (HasRelationshipWithSingularEntity, .ShouldUseRelationshipNameAsNavigation,IsManyRelationshipOnOtherSide) should be part of the EntityRelationship class

That is the only info they need

rochar (changes requested):

Issue #497, opened 2023-08-10
[Nox.Ui] Cryptocash.Ui Phase1 Agree and add Docker support


DanielHodgeGH:

DanielHodgeGH:
https://regusit.atlassian.net/browse/NOX-77

Issue #484, opened 2023-08-09
[Nox.Ui] Cryptocash.Ui add design YAML config once settings agreed


DanielHodgeGH:

DanielHodgeGH:
blocked by #477

PR #498, opened 2023-08-10
feature/481 Initial Commit for Cryptocash.Ui Phase1 Skeleton Layout

DanielHodgeGH:
#481 This PR is for initial skeleton layout of Cryptocash.Ui Phase 1 - so that we can agree how to further slice into template Ui controls for Nox.Generator output

Note:
1 - Yaml config hasnt been added until we agree the Yaml definitions - (example default config has been coded for now into MainLayoutBase) related ticket see #484
2 - PR can be used to agree how to separate out razor into sub Nox.Ui templates for example the Navigation Links and Header sections for now
3 - AzureAd Authorization code has been commented out for now until YAML config agreed (within Program.cs)
4 - Have added Generated folders as an example to be agreed of handling generated content - idea being that the razor page is created (ie: Bookings.razor) and uses the generic page base to handle logic (ie: ApiManagementPageBase in an attempt to reduce Generator output.
5 - Have added a way for child controls to trigger a change in global data  - see ApiManagementPageBase.UpdateHeaderApiName() - to be agreed if this is best approach for child components to trigger and change data globally for other child/parent components with BlazorServer code - see related ticket https://github.com/NoxOrg/Nox.Ui.Sandbox/issues/1
6 - I'll add Docker support under a different ticket see #497
7 - Have left in for now ConfigurationHelper as it is probably useful later on

thanks DanHo


anna-naumova25:
Overall looks good to me. I'd add some small comments below:
1. I would rename ApiManagement folder for generated file to something like Pages. Cause in fact every entity is managed by API call and and it's not ApiManagement itself. We can always discuss it, but I'd suggest something like this:

Generated/Pages/
├── Booking/
│   ├── BookingsList.cshtml
│   ├── BookingCreate.cshtml
│   ├── BookingEdit.cshtml
│   └── BookingDetails.cshtml
├── Employer/
│   ├── EmployersList.cshtml
│   ├── EmployeeCreate.cshtml
│   └── ...
└── ...

2. I would also rename ApiManagementPageBase but couldn't come up with a right name.
3. Maybe it would make sense to rename generated files to *.g.razor, *.g.cs so we can understand what we expect to be generated by Nox and what should Nox users (actually developers) add by themselves. 
4. Based on the previous point, I think it would be good to create a rough README file and put there everything that should be done in the project manually. I know things are going to change many times, but it could be a draft version for now.

rochar (changes requested):

DanielHodgeGH:
#481 PR Requested Changes

1 - As per Anna's request have changed folder structure to /Generated/Pages (plus Components & Data) - might still need discussion if so lets create a meeting to discuss - for example Add and Edit may be modal popup components rather than pages in whole - also have kept root Pages folder to hold index, _host root pages etc to not break blazor
2 - Changed 'Generated' files to include .g.razor or .g.cs
3 - Readme will be added on a separate ticket once Config agreed see #501
4 - Requested changes made to .proj file
5 - Removed extra .sln file
6 - For now have included Customer and Employee entities to tie in with initial Cryptocash.Api project - havnt included Employer as probably safe to assume for this POC only dealing with 1 employer ie: Cryptocash

rochar (changes requested):
looks good, see my comments

rochar (approved):

PR #525, opened 2023-08-14
461 Password property types do not generate in dto

anna-naumova25:
At the moment I've added options for Read and Update only, as Create Dto is generated from Update Dto, and Delete option cannot be implemented as we delete the whole entity not its single property. So Delete option can be added later.


rochar (approved):
Looks good

iDtogenerateControl, probably we can get a more meaningful name
We can discuss this later

PR #527, opened 2023-08-15
MP-8507: NOX - Factory for converting Primitive types into Nox types …

originalethanhunt:
…for "Year"

### DIFF
--- a/src/Nox.Lib/Factories/Types/NoxTypeYearFactory.cs
+++ b/src/Nox.Lib/Factories/Types/NoxTypeYearFactory.cs
@@ -1,0 +1,22 @@
+﻿using Nox.Solution;
+using Nox.Types;
+
+namespace Nox.Factories.Types;
+
+/// <summary>
+/// The nox type year factory.
+/// </summary>
+public class NoxTypeYearFactory : NoxTypeFactoryBase<Year>
+{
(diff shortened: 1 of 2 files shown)


src/Nox.Lib/Factories/Types/NoxTypeYearFactory.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Nox.Factories.Types;
+
+public class NoxTypeYearFactory : NoxTypeFactoryBase<Year>
+{
+    public NoxTypeYearFactory(NoxSolution solution) : base(solution)
+    {
+    }
+
+    public override Year? CreateNoxType(NoxSimpleTypeDefinition simpleTypeDefinition, dynamic? value)
+        => value is null ? null :Year.From(value);
```
dejobratic, line 13: We should be taking options from simpleTypeDefinition, as it was done here https://github.com/NoxOrg/Nox.Generator/blob/e3a689aa33bcce28a07dac21b510e563bfbaf4e8/src/Nox.Lib/Factories/Types/NoxTypeAreaFactory.cs

dejobratic (changes requested):
Looks good. Left a comment.

rochar (approved):

PR #529, opened 2023-08-15
MP-8465 - Factory for converting Primitive types into "Color" 

danield-ap:


src/Nox.Lib/Factories/Types/NoxTypeColorFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Nox.Factories.Types
+{
+    public class NoxTypeColorFactory : NoxTypeFactoryBase<Color>
+    {
+        public NoxTypeColorFactory(NoxSolution solution) : base(solution)
+        {
+        }
+
+        public override Color? CreateNoxType(NoxSimpleTypeDefinition simpleTypeDefinition, dynamic? value)
```
gjorgji-grgovski-authoritypartners, line 12: I think the dynamic value would always come in whatever type we are persisting in the Database, for instance in this case Color is persisted as an array of bytes, so I believe we should only instantiate the Color.From(byte[]).
@rochar will need your confirmation here please.

gjorgji-grgovski-authoritypartners: image

danield-ap: Factory and From method updated.

rochar: @gjorgji-grgovski-authoritypartners they will come with the type defined on the NoxType enum => this will be the type of the dto, so on this case string

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #533, opened 2023-08-15
512 Post with multiple primary keys

anna-naumova25:
I've updated the command to return a record of keys. However we still need changes to support auto generation of keys for both single and multiple key entities.


rochar:
And also, get the actual key values

rochar (approved):

PR #537, opened 2023-08-16
MP-8690 - Auditable Entities Enhancement

dejobratic:
Nox: 
- Added new interfaces IAuditCreated, IAuditUpdated and IAuditDeleted that are implemented by AuditableEntityBase
- Moved all interfaces from Nox.Lib to Nox.Abstractions (including the ones mentioned above)
- Added ISystemProvider and IUserProvider with default implementations that provide required user and system for auditing
- Updated all commands with new ISystemProvider and IUserProvider injected intefaces 
- Changed IsVersioned to IsAudited and updated schemas
- Replaced Deleted flag in Dtos with DeletedAtUtc
- Added definitions for AuditableEntityBase attributes, since all of them are from Nox.Types and they require registering database configurators

SampleWebApp:
 - Refactored data seeders;
 - Updated key for Country entity to be databaseNumber, so that we can test the CRUD endpoints with auditing


src/Nox.Lib/Extensions/ServiceCollectionExtension.cs
```diff
@@ -120,4 +123,13 @@ private static NoxSolution CreateSolution(IServiceProvider serviceProvider)
 
         return services;
     }
+
+    private static IServiceCollection AddNoxProviders(
+        this IServiceCollection services)
+    {
+        services.AddSingleton<IUserProvider, DefaultUserProvider>();
```
rochar: we need to change to scoped in the next PR

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #567, opened 2023-08-16
MP-8521: NOX - Factory for converting Primitive types into Nox types …

originalethanhunt:
…for "DateTimeDuration"


src/Nox.Lib/Factories/Types/NoxTypeDateTimeDurationFactory .cs
dejobratic, line 1: Can we also update the EntityMapper.template file with this type?

dejobratic (changes requested):

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #572, opened 2023-08-16
MP-8535  - Factory for converting Primitive types into "DayOfWeek"

danield-ap:


src/Nox.Lib/Factories/Types/NoxTypeDayOfWeekFactory.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+        }
+
+        public override Nox.Types.DayOfWeek? CreateNoxType(NoxSimpleTypeDefinition simpleTypeDefinition, dynamic? value)
+        {
+            if (value == null)
+            {
+                return null;
+            }
+            return Nox.Types.DayOfWeek.From(value);
```
gjorgji-grgovski-authoritypartners, line 18: DayOfWeek currently is of type "int", can we maybe align the ValueObject primitive type to "ushort" and NoxType.cs to "ushort" 

image

image


@rochar  can you please give your feedback here?

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Lib/Factories/Types/NoxTypeDayOfWeekFactory.cs
rochar: agree, so we consolidate the Value object, the Converter and NoxType

danield-ap: Base type updated.

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #595, opened 2023-08-18
444 noxgenerators odata generate keys that are not computed

### LINKED ISSUE #444: [Nox.Generators] [Odata] Generate keys that are not computed
The generated CreateEntityDto Dto, needs to generate a Property for keys that are not computed(including Entity key) , this means all NoxType except DatabaseNumber, DatabaseGuid, and Nuid.

The user when creating this entity needs to set a value on this keys (set this Property has Required in this model)

anna-naumova25:


src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
```diff
@@ -13,6 +14,18 @@ namespace {{codeGeneratorState.ApplicationNameSpace }}.Dto;
 /// </summary>
 public partial class {{className}} : {{entity.Name}}UpdateDto
 {
-    // TODO Add Manual Keys to be set by the user
-
+{{- for key in entity.Keys }}
+    {{- if key.Type == "Nuid" || key.Type == "DatabaseNumber" || keyType == "Guid" -}}
```
rochar, line 18: should be DatabaseGuid , Guid is not autogenerated

src/Nox.Generator/Application/EntityMapper.template.cs
```diff
@@ -28,6 +28,21 @@ public override void MapToEntity({{entity.Name}} entity, Entity entityDefinition
     #pragma warning disable CS0168 // Variable is declared but never used        
         dynamic? noxTypeValue;
     #pragma warning restore CS0168 // Variable is declared but never used
+    {{ for key in entity.Keys -}}
+    {{- if key.Type == "Nuid" || key.Type == "DatabaseNumber" || keyType == "Guid" -}}
```
rochar, line 32: same here

src/Nox.Lib/Factories/NoxTypeFactoryBase.cs
```diff
@@ -15,8 +15,9 @@ public NoxTypeFactoryBase(NoxSolution solution)
         public virtual T? CreateNoxType(Entity entityDefinition, string propertyName, dynamic? value)
         {
```
rochar, line 16: we should be explicitly and have a method IsKey(keyname) and according to it get the right definition

src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -96,12 +96,27 @@ public NoxSimpleTypeDefinition GetAttributeByName(string entityName)
                 }
             }
         }
-        return _attributesByName[entityName];
+        return _attributesByName.TryGetValue(entityName, out var attribute) ? attribute : null;
```
rochar, line 99: No need to do this, if we follow the previous comment

src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -96,12 +96,27 @@ public NoxSimpleTypeDefinition GetAttributeByName(string entityName)
...
+    public NoxSimpleTypeDefinition? GetKeyByName(string entityName)
+    {
+        lock (this)
+        {
+            if (_keysByName is null && Keys is not null)
+            {
+                _keysByName = Keys.ToDictionary(key => key.Name, key => key);
+            }
+        }
+        return _keysByName?.TryGetValue(entityName, out var key) == true ? key : null;
```
rochar, line 111: the same here

rochar (changes requested):
looks good, comments to review
also I want to show you how we do a end to end test on Nox

alexandervlasenko (approved):
Please, adress the issues raised by Ricardo, otherwise everything is good.

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
anna-naumova25: Updated

src/Nox.Lib/Factories/NoxTypeFactoryBase.cs
anna-naumova25: Added IsKey and updated the codebase accordingly

rochar (approved):

PR #597, opened 2023-08-19
MP-8784: Added the following types to allnoxtype entity

originalethanhunt:
Color
HtmlField
InternetDomainField
IpAddressField
JsonField
JwtTokenField
MonthField
NuidField


src/Nox.Types.Abstractions/Enums/NoxType.cs
```diff
@@ -85,7 +85,7 @@ public enum NoxType : uint
     [SimpleType(typeof(bool))]
     Boolean = 2157507194,
 
-    [SimpleType(typeof(string))]
+    [SimpleType(typeof(byte[]))]
```
rochar, line 88: @gjorgji-grgovski-authoritypartners we will keep Color as string in dtos?

andresharpe: I think we should keep this simple and persist as a string in well known hex format, eg.:

#CD5C5C

andresharpe: so it'll map to varchar(7) in sql server

originalethanhunt: it's ARGB so I guess that will be varchar(9).

rochar: so lets make this string again, so we can merge this PR

@gjorgji-grgovski-authoritypartners  is taking care of creating a tickets to refactor Color Value Object to be string and a converter to varchar(9)

rochar (changes requested):

rochar:
Please also update it from main

rochar (approved):

PR #616, opened 2023-08-22
MP-8556 - Factory for converting Primitive types into "Formula"

danield-ap:


rochar:
formula is a computed field, so no need a mapper to an entity
thanks

PR #629, opened 2023-08-22
MP-8486 Factory for converting Primitive types into Nox types for "CurrencyNu…

msadikkoseaupa:
Also redundant codes were removed from CurrencyNumber implementation


src/Nox.Types/Types/CurrencyNumber/CurrencyNumber.cs
```diff
@@ -10,7 +10,7 @@ namespace Nox.Types;
 [Serializable]
 public sealed class CurrencyNumber : ValueObject<short, CurrencyNumber>
 {
-    private static HashSet<short> _currecyNumbers = new HashSet<short>()
```
dejobratic, line 13: Can we mark this field as readonly?

dejobratic (approved):

src/Nox.Types/Types/CurrencyNumber/CurrencyNumber.cs
msadikkoseaupa: Desired changes were applied

rochar (approved):

PR #632, opened 2023-08-23
MP-8811 [ENHANCE-187] NOX - Refactor Color value object to persist th…

### LINKED ISSUE #635: Refactor Color value object to persist the Hex string value MP-8811

becicajla:
…e Hex string value

### DIFF
--- a/src/Nox.Types/Types/Color/Color.cs
+++ b/src/Nox.Types/Types/Color/Color.cs
@@ -1,21 +1,22 @@
 ﻿using System;
 using System.Drawing;
+using System.Text.RegularExpressions;
 
 namespace Nox.Types;
 
 /// <summary>
 /// Represents a Nox <see cref="Color"/> type and value object. 
 /// </summary>
-public class Color : ValueObject<byte[], Color>
+public class Color : ValueObject<string, Color>
 {
-    private System.Drawing.Color SystemColor { get; set; }
+    private readonly Regex _hexColorRegex = new Regex(@"^#(([A-Fa-f0-9]{3})|([A-Fa-f0-9]{6})|([A-Fa-f0-9]{8}))$");
 
     /// <summary>
     /// Creates a new instance of <see cref="Color"/> object.
     /// </summary>
     public Color()
     {
-        Value = new byte[4];
+        Value = "#00000000";
     }
 
     /// <summary>
@@ -24,7 +25,7 @@ public class Color : ValueObject<byte[], Color>
     /// <param name="color">the color to be parsed.</param>
     public static Color From(System.Drawing.Color color)
     {
-        return From(color.A, color.R, color.G, color.B);
+        return From($"#{color.A:X2}{color.R:X2}{color.G:X2}{color.B:X2}");
     }
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


src/Nox.Types/Types/Color/Color.cs
```diff
@@ -59,10 +60,21 @@ public static Color From(byte alpha, byte red, byte green, byte blue)
     /// <param name="color">The origin value to create the <see cref="Color"/> with</param>
     /// <returns></returns>
     /// <exception cref="TypeValidationException"></exception>
-    public static Color From(string color)
+    public static new Color From(string colorHex)
```
gjorgji-grgovski-authoritypartners: I don't think we need this From method, as it is already part of the base ValueObject.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -74,40 +86,62 @@ public static Color From(string color)
...
+        var color = ToSystemColor();
+        return new[] { color.A, color.R, color.G, color.B };
+    }
 
     /// <summary>
     /// Returns the string representation of the color value, formatted in Hexadecimal format string
     /// </summary>
     /// <returns>The string representation</returns>
-    public string ToHexa() => $"#{SystemColor.A:X2}{SystemColor.R:X2}{SystemColor.G:X2}{SystemColor.B:X2}";
+    public string ToHexa()
```
gjorgji-grgovski-authoritypartners: I don't think we need this method, as it can be accessed through the Value property.

src/Nox.Types/Types/Color/Color.cs
```diff
@@ -74,40 +86,62 @@ public static Color From(string color)
...
+        var color = ToSystemColor();
+        return $"#{color.A:X2}{color.R:X2}{color.G:X2}{color.B:X2}";
+    }
 
     /// <summary>
     /// Returns the string representation of the color value, formatted in Hex format string
     /// </summary>
     /// <returns>The string representation</returns>
-    public string ToHex() => $"#{SystemColor.R:X2}{SystemColor.G:X2}{SystemColor.B:X2}";
+    public string ToHex()
```
gjorgji-grgovski-authoritypartners, line 99: I don't think we need this method, as it can be accessed through the Value property.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types/Types/Color/Color.cs
becicajla: @gjorgji-grgovski-authoritypartners , but color can be presented as only RGB or RGB+alpha (opacity). If we delete toHex or ToHexa, we will lose the conversion to these types. Is that ok?

src/Nox.Types/Types/Color/Color.cs
becicajla: done

src/Nox.Types/Types/Color/Color.cs
becicajla: tohex left, tohexa deleted

src/Nox.Types/Types/Color/Color.cs
becicajla: done

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):

PR #649, opened 2023-08-24
feature 485 Cryptocash Design Yaml Schema for initial review

### LINKED ISSUE #485: [Nox.Ui] Cryptocash.Api add design YAML config once settings agreed

DanielHodgeGH:
Initial PR to review design yaml schema for Cryptocash.Api

1. have added design yaml files in folder structure which matches the solution yaml
2. added small temp test project to generate the mermaid diagram

Found the following issues which need discussion (will retry these with latest Main branch)

1. Formula fails so commented out for yaml for now
2. Guid fails so changed to databaseNumber for now
3. ExchangeRate has 2 separate links to currency entity - need to discuss how to implement
4. Employee entity has issue if isAudit is true - employee might be a reserved name?


DanielHodgeGH:
Discussed with Andre - Formula ok - was just missing the formula definition

the following we think are bugs:
1. Guid fails so changed to databaseNumber for now
2. ExchangeRate has 2 separate links to currency entity - need to discuss how to implement
3. Employee entity has issue if isAudit is true - employee might be a reserved name?

cc @rochar

PR #652, opened 2023-08-24
568 Post To Owned Entity in Parent Entity Odata Controller

### LINKED ISSUE #568: [Nox.Generators] Post/Put/Patch To Owned Entity in Parent Entity Odata Controller (Many)
Investigate the best practice for this, should the Create<EntityDto> receive single/List of key(s) to the related entity?

For now let's supports single key for related entities

anna-naumova25:
This is an initial commit for #568. Please do **not** close the ticket and the branch after merge.

- AddOwnedEntityCommand has been added.
- EntityKeyDto has been transformed to a parent class for EntityDto because of the OData exception below. Can discuss it later on the call if need.

`The type 'SampleWebApp.Application.Dto.CountryLocalNameKeyDto' does not inherit from and is not a base type of 'SampleWebApp.Application.Dto.CountryLocalNameDto'. The type of 'KeySegments' must be related to the Type of the EntitySet.'`

- Will continue working on Post for parent entity to add owned entities list.
- Will add more tests to cover OwnedEntities scenraios.


src/Nox.Generator/Application/Commands/AddCommand.template.cs
```diff
@@ -0,0 +1,64 @@
...
+		var entity = EntityFactory.CreateEntity(request.EntityDto);
+
+		{{- for key in entity.Keys ~}}
+		{{- if key.Type == "Nuid" }}
+		entityToCreate.Ensure{{key.Name}}();
+		{{- end }}
+		{{- end }}
+
+		parentEntity.{{entity.PluralName}}.Add(entity);
+
```
rochar, line 61: add OnCompleted

src/SampleWebApp/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Dto.CountryDto.g.cs
```diff
@@ -11,18 +12,21 @@
...
-/// The list of countries.
-/// </summary>
-public partial class CountryDto
+public class CountryKeyDto
 {
 
     /// <summary>
     ///  (Required).
     /// </summary>
+    [Key]
```
rochar: we need to understand why? and use NoxDtoDatabaseConfigurator to do this

rochar (changes requested):

src/Nox.Generator/Application/Commands/AddCommand.template.cs
anna-naumova25: Updated

src/SampleWebApp/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Dto.CountryDto.g.cs
anna-naumova25: As discussed on the call, get rid of the Key attribute, leave EntityKeyDto as a Record and return ChildDto with keys only from PostToOwnedEntity method

rochar (approved):

PR #661, opened 2023-08-25
Feature/660 Parent Post includes Owned List

### LINKED ISSUE #660: [Nox.Generators] Put/Patch To Owned Entity in Parent Entity Odata Controller (Many)
BASED ON #568 

1. Make sure that owned entities are added successfully when they are included to parent Post request.
2. Put/Patch for an owned entity, meaning the parent entity already exists, example:
Example
PUT http://localhost:5000/odata/Employees(4)/Peers
We should document this in our readme and define the workflow for delete a owned entity

@rochar FYI

anna-naumova25:


src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
```diff
@@ -37,9 +37,9 @@ public partial class {{className}}
     /// {{entity.Name}} {{relationship.Description}} {{relationship.Relationship}} {{relationship.EntityPlural}}
     /// </summary>
     {{- if relationship.Relationship == "ZeroOrMany" || relationship.Relationship == "OneOrMany"}}
-    public virtual List<{{relationship.Entity}}UpdateDto> {{relationship.EntityPlural}} { get; set; } = new();
+    public virtual List<{{relationship.Entity}}Dto> {{relationship.EntityPlural}} { get; set; } = new();
     {{- else}}
-     public virtual {{relationship.Entity}}UpdateDto{{- if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Entity}} { get; set; } = null!;
+    public virtual {{relationship.Entity}}Dto{{- if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Entity}} { get; set; } = null!;
```
rochar, line 42: Why this change is needed?, the owned in the update should also be an updateDto? for example we can not set autogenerated ids?

tests/Nox.ClientApi.Tests/Tests/Controllers/CountriesControllerTests.cs
```diff
@@ -67,7 +67,7 @@ public async Task Post_WithManyOwnedEntity_ReturnsDatabaseNumberId(ApiFixture ap
                 new CountryCreateDto
                 {
                     Name = apiFixture.Fixture.Create<string>(),
-                    CountryLocalNames = new List<CountryLocalNameUpdateDto>() { new CountryLocalNameUpdateDto() { Name = expectedOwnedName } }
+                    CountryLocalNames = new List<CountryLocalNameDto>() { new CountryLocalNameDto() { Name = expectedOwnedName } }
```
rochar, line 70: yup we need to discuss why using the Dto instead of the update dto?

tests/Nox.ClientApi.Tests/Tests/Controllers/CountriesControllerTests.cs
```diff
@@ -79,9 +79,8 @@ public async Task Post_WithManyOwnedEntity_ReturnsDatabaseNumberId(ApiFixture ap
                 .Which.Entity.keyId.Should().BeGreaterThan(0);
 
             queryResult.Should().NotBeNull();
-
-            // TODO: add odata controller to test include properly
-            //queryResult!.ToDto().OwnedEntities!.Single().Name.Should().Be(expectedOwnedName);
+            queryResult.ExtractResult().CountryLocalNames.Should().NotBeNull();
```
rochar, line 82: I'm curious, not sure how you were able to retrieve/load the owned entity not having Odata working and set AutoExpand in this test project?

rochar (changes requested):
looks good, I have one doubt regarding not using owned UpdateDto

tests/Nox.ClientApi.Tests/Tests/Controllers/CountriesControllerTests.cs
anna-naumova25: I don't know the reason why it didn't work before, but now it works at this test runs successfully

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
anna-naumova25: We can discuss it later on a call, I'll put some points why changed it:

- in the scenario for this PR I needed CreateEntityDto model for Owned entities. However CreateEntityDto is inherited from UpdateEntityDto model, so I have change parent UpdateEntityDto;
- putting CreateEntityDto children into UpdateEntityDto parent doesn't look good to me;
-  in fact the UpdateEntityDto model for owned entities is useless cause it doesn't contain Ids and we need to pass the children's Ids to update owned entities as we can update them from the parent controller only;
- so for now I put EntityDto for owned entities into the UpdateEntityDto parent model which seems to cover all the cases, we can always discuss and change it if needed.

andresharpe:
Can you resolve the conversations please Anna?

anna-naumova25:
sure @andresharpe, I've resolved the conversations

alexandervlasenko (approved):

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
rochar: Thanks, let then talk about this next monday. Can you please update with the latest from main, so it's easier to understand the changes,

anna-naumova25: I've updated the final PR that includes all the changes for this ticket - https://github.com/NoxOrg/Nox.Generator/pull/673

PR #674, opened 2023-08-29
Feature/#662 implement relationship mechanic in generated dto context

ILoveMyTortoise:
Sorry for a big one


alexandervlasenko (approved):

AndreyDegtyarev (approved):

.nox/design/store-owner.entity.nox.yaml
```diff
@@ -0,0 +1,32 @@
+# store.entity.nox.yaml
+
+name: StoreOwner
+
+description: Store owners
+
+keys:
+
+  - name: Id
+    isRequired: true
```
gjorgji-grgovski-authoritypartners: Let's move the "isRequired" property at the bottom, just for consistency.

tests/Nox.ClientApi.Tests/.nox/design/Entitities/store-owner.entity.nox.yaml
```diff
@@ -0,0 +1,32 @@
+# store.entity.nox.yaml
+
+name: StoreOwner
+
+description: Store owners
+
+keys:
+
+  - name: Id
+    isRequired: true
```
gjorgji-grgovski-authoritypartners: Let's move the "isRequired" property at the bottom, just for consistency.

src/Nox.Types.EntityFramework/Configurations/NoxDtoDatabaseConfigurator.cs
```diff
@@ -43,6 +45,43 @@ private static void ConfigureAttributes(NoxSolutionCodeGeneratorState codeGenera
             }
         }
     }
+
+    private static void ConfigureRelations(NoxSolutionCodeGeneratorState codeGeneratorState, IEntityBuilder builder, Entity entity)
```
gjorgji-grgovski-authoritypartners: This is the same code as in the NoxDatabaseConfigurator, I would suggest in the next iteration to make some refactoring and avoid duplication of a code.
CC: @rochar

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
```diff
@@ -26,11 +26,20 @@ public partial class {{className}}
...
+    public {{attribute.Type}}Dto{{- if !attribute.IsRequired}}?{{end}} {{attribute.Name}} { get; set; }{{if attribute.IsRequired}} = default!;{{end}}
     {{- end}}
 {{- end }}
+{{- for relationship in entity.Relationships}}
+    {{- if relationship.Relationship == "ZeroOrOne" || relationship.Relationship == "ExactlyOne" && relationship.ShouldGenerateForeignOnThisSide}}
+
+    /// <summary>
+    /// {{entity.Name}} {{relationship.Description}} {{relationship.Relationship}} {{relationship.EntityPlural}}
+    /// </summary>
+    public string{{if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Entity}}Id { get; set; } = null!;
```
gjorgji-grgovski-authoritypartners: The type of the relationship property has to vary based on the entity key primitive type.

src/Nox.Generator/Application/Dto/EntityDto.template.cs
```diff
@@ -54,9 +54,9 @@ public partial class {{className}}
     {{- else}}
         {{- if relationship.ShouldGenerateForeignOnThisSide}}
     //EF maps ForeignKey Automatically
-    public virtual string {{if relationship.Relationship == "ZeroOrOne"}}?{{end}}{{relationship.Entity}}Id { get; set; } = null!;
+    public string{{if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Entity}}Id { get; set; } = null!;
```
gjorgji-grgovski-authoritypartners: The type of the relationship property has to vary based on the entity key primitive type.

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
gjorgji-grgovski-authoritypartners: Relationships like oneOrMany or zeroOrMany are not generating properties in the Dto, is that expected or not?

gjorgji-grgovski-authoritypartners: Do we need to have a Required data annotation attribute on the properties for the relationships like exactlyOne, oneOrMany?

src/Nox.Generator/Application/Dto/EntityDto.template.cs
gjorgji-grgovski-authoritypartners: Do we need to have a Required data annotation attribute on the properties for the relationships like exactlyOne, oneOrMany?

src/Nox.Generator/Application/Dto/EntityDto.template.cs
```diff
@@ -54,9 +54,9 @@ public partial class {{className}}
...
         {{- end}}
-    public virtual {{relationship.Entity}}Dto {{if relationship.Relationship == "ZeroOrOne"}}?{{end}}{{relationship.Entity}} { get; set; } = null!;
+    public virtual {{relationship.Entity}}Dto{{if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Entity}} { get; set; } = null!;
```
gjorgji-grgovski-authoritypartners, line 59: Why do we have a virtual keyword for the properties in the Dto?

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
```diff
@@ -46,29 +46,45 @@ public override void MapToEntity({{entity.Name}} entity, Entity entityDefinition
...
+        {        
+            entity.{{relationship.Entity}}Id = noxTypeValue;
+        }
+    {{-end}}
     {{- end }}
     }
 
     public override void PartialMapToEntity({{entity.Name}} entity, Entity entityDefinition, Dictionary<string, dynamic> updatedProperties)
     {
+#pragma warning disable CS0168 // Variable is assigned but its value is never used
```
gjorgji-grgovski-authoritypartners, line 76: We don't need to declare this variable.

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
```diff
@@ -84,5 +100,22 @@ public override void PartialMapToEntity({{entity.Name}} entity, Entity entityDef
...
         }
     {{- end }}
+    
+    {{ for relationship in entity.Relationships }}
+
+    {{- if (relationship.Relationship == "ZeroOrOne" || relationship.Relationship == "ExactlyOne") && relationship.ShouldGenerateForeignOnThisSide}}
+        /// <summary>
+        /// {{entity.Name}} {{relationship.Description}} {{relationship.Relationship}} {{relationship.EntityPlural}}
+        /// </summary>
+        if (updatedProperties.TryGetValue("{{relationship.Entity}}Id", out value))
```
gjorgji-grgovski-authoritypartners, line 110: Can we call the output parameter something like {{relationship.Entity}}IdValue?

If it is complicated to achieve, because by naming convention should be camelCase, let's keep it like this for now and refactor it later.

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
```diff
@@ -46,29 +46,45 @@ public override void MapToEntity({{entity.Name}} entity, Entity entityDefinition
...
+    {{- end }}
+    {{- end }}
+    {{ for relationship in entity.Relationships }}
+    {{- if (relationship.Relationship == "ZeroOrOne" || relationship.Relationship == "ExactlyOne") && relationship.ShouldGenerateForeignOnThisSide}}
+
+        /// <summary>
+        /// {{entity.Name}} {{relationship.Description}} {{relationship.Relationship}} {{relationship.EntityPlural}}
+        /// </summary>
+        noxTypeValue = CreateNoxType<Nox.Types.{{relationship.Related.Entity.Keys[0].Type}}>(entityDefinition, "{{relationship.Name}}", dto.{{relationship.Entity}}Id);
+        if (noxTypeValue != null)
```
gjorgji-grgovski-authoritypartners, line 66: In this case value of the relationship entity ID should be populated/required, so we might need to throw an Exception if the noxTypeValue is null.

src/Nox.Lib/Factories/NoxTypeFactoryBase.cs
gjorgji-grgovski-authoritypartners, line 1: We can also just extend Entity.cs with methods "IsAttribute" and "IsRelationship" and based on that call proper "Get{Whatever}ByName" method.

`NoxSimpleTypeDefinition? definition;
 if (entityDefinition.IsKey(propertyName))
                definition = entityDefinition.GetKeyByName(propertyName);
 else if (entityDefinition.IsAttribute(propertyName))
                definition = entityDefinition.GetAttributeByName(propertyName);
 else if (entityDefinition.IsRelation(propertyName))
                definition = entityDefinition.GetRelationByName(propertyName);`

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
```diff
@@ -26,11 +26,20 @@ public partial class {{className}}
...
     {{ if componentsInfo[attribute.Name].IsSimpleType -}}
-    public {{componentsInfo[attribute.Name].ComponentType}}{{ if !attribute.IsRequired}}?{{end}} {{attribute.Name}} { get; set; } {{if attribute.IsRequired}}= default!;{{end}}
+    public {{componentsInfo[attribute.Name].ComponentType}}{{ if !attribute.IsRequired}}?{{end}} {{attribute.Name}} { get; set; }{{if attribute.IsRequired}} = default!;{{end}}
     {{- else -}}
-    public {{attribute.Type}}Dto{{- if !attribute.IsRequired}}?{{end}} {{attribute.Name}} { get; set; } {{if attribute.IsRequired}}= default!;{{end}}
+    public {{attribute.Type}}Dto{{- if !attribute.IsRequired}}?{{end}} {{attribute.Name}} { get; set; }{{if attribute.IsRequired}} = default!;{{end}}
     {{- end}}
 {{- end }}
+{{- for relationship in entity.Relationships}}
+    {{- if relationship.Relationship == "ZeroOrOne" || relationship.Relationship == "ExactlyOne" && relationship.ShouldGenerateForeignOnThisSide}}
```
gjorgji-grgovski-authoritypartners:
```suggestion
    {{- if relationship.WithSingleEntity && relationship.ShouldGenerateForeignOnThisSide}}
```

src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -116,6 +122,33 @@ public virtual bool IsKey(string keyName)
...
+    {
+        result = null;
+        if (Relationships == null)
+        {
+            return false;
+        }
+
+        var rel = Relationships!.First(x => x.Name.Equals(relationshipName));
+        // TODO: possibly extend for other types
+        if (!rel.ShouldGenerateForeignOnThisSide ||
```
gjorgji-grgovski-authoritypartners, line 135:
```suggestion
        if (!rel.ShouldGenerateForeignOnThisSide || rel.IsManyRelationshipOnOtherSide)
```

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
gjorgji-grgovski-authoritypartners, line 1: Why don't we have the attributes on the CreateDtos?

tests/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData.ODataServiceCollectionExtensions.g.cs
```diff
@@ -46,6 +47,12 @@ public static void AddNoxOdata(this IServiceCollection services)
         builder.EntityType<WorkplaceDto>();
         builder.EntityType<WorkplaceKeyDto>();
 
+        builder.EntitySet<StoreOwnerDto>("StoreOwners");
+
+        builder.EntityType<StoreOwnerDto>();
+        builder.EntityType<StoreOwnerKeyDto>();
+        builder.EntityType<StoreOwnerDto>().Ignore(e => e.DeletedAtUtc);
```
gjorgji-grgovski-authoritypartners, line 54: Instead of ignoring the DeleteAtUtc property, why don't we just remove it from the Dto if we don't need it?

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Types.EntityFramework/Configurations/NoxDtoDatabaseConfigurator.cs
ILoveMyTortoise: We can discuss this with Ricardo, not sure if this is an optimal solution, because it actually uses different fields and will require a lot of params, also we may need to change them independently

src/Nox.Generator/Application/Dto/EntityDto.template.cs
ILoveMyTortoise: It's only for navigation properties, we are using EF to operate them in dtos and marking them as virtual is a good practice

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
ILoveMyTortoise: we need, because if we don't, we have to declare it in out parameter, and if so, we will get "Error	CS0136	A local or parameter named 'value' cannot be declared in this scope because that name is used in an enclosing local scope to define a local or parameter"

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
ILoveMyTortoise: Because we have them in UpdateDto and inherit from it, to not duplicate it

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
ILoveMyTortoise: will fix it in couple other places too

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
ILoveMyTortoise: No, because foreign keys according to EF notation are named {Entity}Id

tests/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData.ODataServiceCollectionExtensions.g.cs
ILoveMyTortoise: I'm not 100% sure if this can be removed, will need to discuss it with @rochar afterwards

src/Nox.Lib/Factories/NoxTypeFactoryBase.cs
ILoveMyTortoise: but why? It will be worse from the performance standpoint and is not really changing the code structure

src/Nox.Solution/Models/Domain/Entities/Entity.cs
ILoveMyTortoise: Actually, this property is about a different thing, but I will add a WithMultiEntity property with the logic you are expecting

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
ILoveMyTortoise: > Relationships like oneOrMany or zeroOrMany are not generating properties in the Dto, is that expected or not?

Yes, they don't need a foreign key

ILoveMyTortoise: I don't think that we need to add this attribute since we currently don't have validation on create at all, even for required entity properties. We have validation during partial update in the PartialMapToEntity in mappers. In the same time, we don't have it yet in MapToEntity. I think validation should be implemented on this level for all fields, including navigation properties, but this should be out of the scope of this task.
@rochar should we create a task for this?

src/Nox.Generator/Application/Dto/EntityDto.template.cs
ILoveMyTortoise: > Do we need to have a Required data annotation attribute on the properties for the relationships like exactlyOne, oneOrMany?

Replied above

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
ILoveMyTortoise: replied above. Yes, but we don't have this for keys and attributes right now, so I think we need to have a task for implementing this, including validating that a related entity exists when specified.

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
ILoveMyTortoise: > The type of the relationship property has to vary based on the entity key primitive type.

Good catch. Actually no, they shouldn't because in our factories almost everywhere we expect a string and have logic to parse it from a string. In this DTO we expect string input from the user and later we convert it into nox type. But I found one small exception, that can be discussed with @rochar later, but doesn't impact the whole flow right now.

src/Nox.Generator/Application/Dto/EntityDto.template.cs
ILoveMyTortoise: > The type of the relationship property has to vary based on the entity key primitive type.

Replied above

ILoveMyTortoise: Done

gjorgji-grgovski-authoritypartners (approved):

PR #677, opened 2023-08-30
Feature/static dto entity mapping

andresharpe:
- Removed need for 'dynamic' and runtime type factoties
- Added static methods to entities to encapsulate defined TypeOptions for each property
- Added static methods to entities to create properties with defined TypeOptions
- Added a ToEntity() method on all OData dto's that creates and maps the dto to an entity
- TODO: refactor key setter to better handle NUID keys, but still handle key setting


alexandervlasenko:
Hi @andresharpe ,
After switching to you branch I've got the following problem with Money options in the generated code:
`MinValue = -999999999,9999m,
MaxValue = 999999999,9999m,`
because 999999999.9999 transformed into 999999999,9999 in my local culture.

I changed the generation code here :
`else if (value is decimal decValue)
{
    return $"{decValue.ToString(CultureInfo.InvariantCulture)}m";
}`
and it fixed the issue. We may experience similar issues in other places because there is no warning about specifying culture info in string interpolation case. Maybe we need to enforce invariant culture info for the generators.

alexandervlasenko (approved):

andresharpe:
Thanks Oleks - have updated the ToSourceCode() to properly handle fractional types (decimal, float, double)

rochar (approved):

PR #685, opened 2023-08-31
Post/Put/Delete Entity with relation to another Entity

anna-naumova25:
- Added POST and PUT to create reference to related entity
- Added DELETE to remove reference to related entity


rochar:
closing it by now until we design the Relations EndPoints

PR #687, opened 2023-09-01
Issues/637 ambigious entity name

### LINKED ISSUE #637: [Nox.Generators] If an Entity is called User or Application, the compilation fails
the same happens if a key is of type guid

AndreyDegtyarev:
_/*Most changes are related to code generator changes
The only things that have been changed are : templates files (like DeleteByIdCommand.template) and added PropertyNameValidator
*/_

Added namespace prefix for generated entity to avoid ambiguity.
Added validator to compare property name and class name.


alexandervlasenko (approved):

src/Nox.Solution/Validation/Domain/EntityValidator.cs
```diff
@@ -60,23 +66,10 @@ public EntityValidator(IEnumerable<Entity>? entities, Application? application)
             RuleForEach(e => e.Events)
                 .SetValidator(e => new DomainEventValidator(domainEvents, appEvents, e.Name));
 
-            RuleForEach(p => p.Keys)
```
ILoveMyTortoise, line 63: Why remove?

ILoveMyTortoise (approved):

rochar (approved):

src/Nox.Solution/ValidationResources.resx
```diff
@@ -453,4 +453,7 @@
   <data name="EntityUniqueAttributeConstraintCanReferenceOnlyExistingAttributes" xml:space="preserve">
     <value>Attribute name '{0}' in unique attribute constraint not found in neither entity attribute(s) ({1}) nor entity key(s) ({2})</value>
   </data>
+  <data name="NameSameAsEnclosingType" xml:space="preserve">
+    <value>{0}. The member {1} of a class cannot have the same name as the class or struct, unless the member is a constructor.</value>
```
ILoveMyTortoise: The member {1} of a class {0} cannot have the same name as the referenced entity.

PR #699, opened 2023-09-04
Added auditability and sorting entities in docs by ownership

andresharpe:


src/Nox.Docs/Extensions/StringExtensions.cs
rochar, line 1: Feels that we need a Nox.Common to have all this extensions :) (Not ideal, but I do not have a better solution)

rochar (approved):

PR #711, opened 2023-09-05
MP-8919 - Optimistic Concurrency with Entity Tag and OData

### LINKED ISSUE #664: Optimistic Concurrency with Entity Tag and OData MP-8919

emir-sehovic1:


src/Nox.Abstractions/Domain/IConcurrent.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Nox.Domain;
+
+/// <summary>
+/// Entities whose updates need to be handled with optimistic concurrency
+/// </summary>
+public interface IConcurrent
```
rochar:
```suggestion
public interface IEntityConcurrent
```

src/Nox.Generator/Presentation/Api/OData/EntityControllerGenerator.cs
```diff
@@ -460,4 +497,21 @@ private static string PrimaryKeysQuery(Entity entity, string prefix = "key")
             string.Join(", ", entity.Keys.Select(k => $"{prefix}{k.Name}")) :
             $"{prefix}";
     }
+
+    private static void GeneratePrivateMethods(CodeBuilder code)
```
rochar: This method can be on an Extension on Nox.Lib, its not necessary to generate code for it

src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -100,4 +100,11 @@ public partial class {{className}}{{ if !entity.IsOwnedEntity }} : {{if entity.P
      public virtual {{relationship.Entity}}{{if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Entity}} { get; set; } = null!;
     {{-end}}
 {{- end }}
+
+{{ if !entity.IsOwnedEntity ~}}
+    /// <summary>
+    /// Entity tag used as concurrency token.
+    /// </summary>
+    public Nox.Types.Guid Etag { get; set; } = Nox.Types.Guid.NewGuid();
```
rochar: you should use the primitive type Guid, there is no added value on being a NoxType
And also will make easy the databaseconfigurator

src/Nox.Abstractions/Domain/IConcurrent.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    /// <summary>
+    /// Gets the entity tag.
+    /// </summary>
+    Types.Guid Etag { get; }
```
rochar: you should use the primitive type Guid, there is no added value on being a NoxType (we are doing the same change for IAuditable...)
And also will make easy the databaseconfigurator

src/Nox.Generator/Application/Dto/EntityDto.template.cs
```diff
@@ -77,5 +77,10 @@ public partial class {{className}}
 
 {{- if !entity.IsOwnedEntity && entity.Persistence?.IsAudited == true}}
     public System.DateTime? DeletedAtUtc { get; set; }
-{{- end }}    
+{{- end }}
+{{- if !entity.IsOwnedEntity }}
+
+    [JsonPropertyName("@odata.etag")]
+    public System.Guid Etag { get; set; }
```
rochar:
```suggestion
    public System.Guid Etag { get; private set; }
```

src/Nox.Types.EntityFramework/Configurations/ConcurrentEntityConfiguration.cs
rochar, line 1: using the primitive type guid , this class should not be needed

src/Nox.ClientApi.Tests/Tests/Application/Commands/CreateCountryCommandHandlerTests.cs
```diff
@@ -41,7 +41,10 @@ public async Task Put_PopulationNegative_ShouldUpdateTo0()
         // Act
 
         var result = await _oDataFixture.PostAsync<CountryCreateDto, CountryKeyDto>(CountryControllerName, countryDto);
-        await _oDataFixture.PutAsync($"{CountryControllerName}/{result!.keyId}", countryUpdateDto);
+        var etag = await GetEtagAsync(result);
```
rochar: meanwhile we change the post to return the full entity, so synchronize main and the result will be the entittydto with the header?

src/Nox.ClientApi.Tests/Tests/Controllers/ODataResponse.cs
```diff
@@ -1,12 +1,16 @@
...
+using System.Text.Json.Serialization;
+
 namespace Nox.ClientApi.Tests.Tests.Controllers
 {
     public class ODataResponse<T>
     {
         public T Value { get; set; } = default!;
+
         [JsonProperty("@odata.count")]
+        [JsonPropertyName("@odata.count")]
```
rochar: why do we need both? is it because System.Text? can we use newton, because its being used on the other tests and work by default better then sytem.text

or remove Newtonsoft.Json; completely

src/Nox.ClientApi.Tests/Tests/NoxIntegrationTestBase.cs
```diff
@@ -53,28 +57,67 @@ public async Task<HttpResponseMessage> PostAsync<TValue>(string requestUrl, TVal
...
+        return message;
+    }
+
+    public Dictionary<string, IEnumerable<string>>? CreateEtagHeader(System.Guid? etag)
+        => new()
+        {
+                { "If-Match", new List<string> { $"\"{etag}\"" } }
+        };
+
+    private static void AddHeaders(HttpClient httpClient, Dictionary<string, IEnumerable<string>>? headers)
```
rochar: we should remove the nullable ?, and avoid if null checks, because collections should not be null but empty

who calls the method needs to take care of this

rochar (changes requested):
looks good, jus a few comments

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #722, opened 2023-09-06
Issues/655 relationship keys validation

### LINKED ISSUE #655: [Nox.Solution] Add Keys Validation
Until now keys were mandatory, we have remove this and now we need to add the validation with the following rules:

1. Owned entity with ZeroOrOne or ExactlyOne relationship keys must be null (See Nox.ClientApi.Test VerifiedEmail Entity as an example)
2. Keys are mandatory for all other entity usages
3. An Entity with a Key to Another entit, type: entityId (act as Foreign key), the dependant entity can only have a single key (not a composite key) See CurrencyCashBalance from CryptoCash as an example) - This is a Nox limitation for now

AndreyDegtyarev:
Provided validation rules for relationships.
Introduced validator EntityRelationshipKeysValidator


src/Nox.Solution/ValidationResources.resx
```diff
@@ -459,4 +459,13 @@
   <data name="EntityOwnedCannotBeAuditable" xml:space="preserve">
     <value>The owned entity '{0}' cannot be auditable.</value>
   </data>
+  <data name="EntityKeysRequired" xml:space="preserve">
+    <value>Keys are mandatory for entity {0} if only it's not Owned entity with ZeroOrOne or ExactlyOne relationship keys.</value>
+  </data>
+  <data name="OwnedEntityKeysMustBeNull" xml:space="preserve">
+    <value>Owned entity {0} with ZeroOrOne or ExactlyOne relationship keys must be null</value>
```
rochar:
```suggestion
    <value>Owned entity {0} with ZeroOrOne or ExactlyOne relationship can not have key(s).</value>
```

src/Nox.Solution/ValidationResources.resx
```diff
@@ -459,4 +459,13 @@
   <data name="EntityOwnedCannotBeAuditable" xml:space="preserve">
     <value>The owned entity '{0}' cannot be auditable.</value>
   </data>
+  <data name="EntityKeysRequired" xml:space="preserve">
+    <value>Keys are mandatory for entity {0} if only it's not Owned entity with ZeroOrOne or ExactlyOne relationship keys.</value>
```
rochar:
```suggestion
    <value>Keys are mandatory for entity {0}. Except owned entities with ZeroOrOne or ExactlyOne relationships.</value>
```

src/Nox.Solution/Validation/Domain/EntityValidator.cs
```diff
@@ -34,6 +32,10 @@ public EntityValidator(IEnumerable<Entity>? entities, Application? application)
                 .SetValidator(x => new EntityItemUniquenessValidator<EntityRelationship>(x, t => t.Name, nameof(x.OwnedRelationships)))
                 .SetValidator(e => new UniquePropertyValidator<EntityRelationship>(e.OwnedRelationships, x => x.Name, "entity owned relation"));
 
+            RuleForEach(e => e.OwnedRelationships)
```
rochar, line 35: should apply to OwnedRelationships and Relationships

rochar (changes requested):
looks good, just some comments

src/Nox.Solution/Validation/Domain/EntityValidator.cs
AndreyDegtyarev: Applied for Relationships as well, but it's not shown in comment differences.

rochar (approved):

PR #727, opened 2023-09-06
MP-9062  - StreetAddress if Required, all properties are required

dejobratic:
- Updated StreetAddressDatabaseConfigurator by marking AddressLine1, PostalCode and CountryId as required. Other properties are marked as optional.
- Updated IStreetAddress, StreetAddress and StreetAddress according to the requred and optional properties. Optional properties are marked as nullable.
- Extended CompoundComponent attribute with IsNullable flag.
- Updated the NoxTypeDtoGenerator logic to take the CompoundComponent nullable flag into consideration when generating the DTOs.


src/Nox.Types.Abstractions/Attributes/Types/CompoundComponentAttribute.cs
```diff
@@ -7,10 +7,12 @@ public class CompoundComponent : Attribute
 {
     public string Name { get; }
     public Type UnderlyingType { get; }
-    public CompoundComponent(string name, Type type)
+    public bool IsNullable { get; set; }
```
rochar: maybe would be more meaningful if we rename to IsRequired?

src/Nox.Types.EntityFramework/Types/StreetAddress/StreetAddressDatabaseConfigurator.cs
```diff
@@ -22,29 +21,47 @@ public class StreetAddressDatabaseConfigurator : INoxTypeDatabaseConfigurator
             x =>
             {
                 x.Ignore(nameof(StreetAddress.Value));
+
```
rochar, line 29: we could get the compoundtype attribute and use it to define the IsRequired so we do not need to change in two places... because we will forget, do you think the effort is worth?

src/Nox.ClientApi.Tests/Tests/Controllers/StoreOwnersControllerTests.cs
```diff
@@ -94,5 +94,40 @@ public async Task Post_VatNumberIsCreated()
             queryResult.Should().NotBeNull();
             queryResult!.VatNumber!.Number.Should().Be(expectedVatNumber);
         }
+
+        [Fact]
+        public async Task Post_StreetAddressIsCreated()
```
rochar: I think we can add two additional tests:

1 - Asserting that we have an exception if we do not fill for example the Country, or required field
2 - Asserting it succeeds if only mandatory fields are set

```suggestion
        public async Task Post_StreetAddressAllFieldsSet_IsCreated()
```

rochar (changes requested):
looks good, see my comments

src/Nox.Types.Abstractions/Attributes/Types/CompoundComponentAttribute.cs
dejobratic: Yes, you are right. Changes were applied.

src/Nox.Types.EntityFramework/Types/StreetAddress/StreetAddressDatabaseConfigurator.cs
dejobratic: Yes, I agree. Changes were applied.

src/Nox.ClientApi.Tests/Tests/Controllers/StoreOwnersControllerTests.cs
dejobratic: Additional tests were added.

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #729, opened 2023-09-06
724 Return OwnedEntityDto from PostTo, PutTo, PatchTo

### LINKED ISSUE #724: Return OwnedEntityDto from PostTo, PutTo, PatchTo
Issue 5 from #611

### Steps to reproduce

1. Add ZeroOrMany relationship to Country entity in SampleWebApp
```
ownedRelationships:
  - name: CountryLocalNameRel
    description: is also know as
    relationship: zeroOrMany
    entity: CountryLocalName
``` 
2. Identify CountryLocalName entity 
```
- name: CountryLocalName
  description: The name of a country in other languages

  persistence:
    isAudited: false

  keys:
    - name: Id
      isRequired: true
      type: databaseNumber

  attributes:
    - name: Name
      isRequired: true
      type: text
      textTypeOptions:
        minLength: 5
        maxLength: 10

    - name: NameCode
      isRequired: false
      type: countryCode2
    
    - name: CustomNumber
      isRequired: false
      type: number
      numberTypeOptions:
       minValue: 100
       maxValue: 1000
```  
3. POST /api/Countries/4/CountryLocalNames
```
{
  "name": "clncln2",
  "nameCode": "FR"
}
``` 
### Actual result

- new CountryLocalName is added to the existing list of CountryLocalNames for Country(4)
- the exception is thrown for the return type CountryLocalNameDto
```
can't parse JSON.  Raw result:

{"@odata.context":"https://localhost:7024/api/$metadata#Countries(10)/CountryLocalNames/$entity","Id":11{"statusCode":500,"message":"Error occured during request: /api/Countries/10/CountryLocalNames.Error: The property \u0027Name[Nullable=False]\u0027 of type \u0027Edm.String\u0027 has a null value, which is not allowed."}
``` 
If happens because we're trying to return CountryLocalNameDto model with only Keys set and OData validation fails for the required properties (Name in this case). This will also reproduce for PatchToCountryLocalName
Need to discuss whether we want to return to CountryLocalNameDto inheritance from CountrlyLocalNameKeyDto to fix it OR if we make another get call to retrun the full CountryLocalNameDto model.
### Expected result
CountryLocalName should be added successfully with the corresponding response

### Solution
Return the OwnedEntityDto (do an extra query), so we are compliant with Odata

anna-naumova25:
- Return OwnedEntityDto from PostTo, PutTo, PatchTo
- Add route attributes for PutTo and PatchTo based on Odata documentation https://learn.microsoft.com/en-us/odata/webapi/attribute-routing

### DIFF
--- a/src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData.CountriesController.g.cs
+++ b/src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData.CountriesController.g.cs
@@ -83,7 +83,7 @@ public partial class CountriesController : ODataController
     }
     
     [HttpPut("/api/Countries/{key}/CountryLocalNames/{relatedKey}")]
-    public async Task<ActionResult> PutToCountryLocalNamesNonConventional( System.Int64 key,  System.Int64 relatedKey, [FromBody] CountryLocalNameUpdateDto countryLocalName)
+    public async Task<ActionResult> PutToCountryLocalNamesNonConventional(System.Int64 key, System.Int64 relatedKey, [FromBody] CountryLocalNameUpdateDto countryLocalName)
     {
         if (!ModelState.IsValid)
         {
@@ -106,7 +106,7 @@ public partial class CountriesController : ODataController
     }
     
     [HttpPatch("/api/Countries/{key}/CountryLocalNames/{relatedKey}")]
-    public async Task<ActionResult> PatchToCountryLocalNamesNonConventional( System.Int64 key,  System.Int64 relatedKey, [FromBody] Delta<CountryLocalNameUpdateDto> countryLocalName)
+    public async Task<ActionResult> PatchToCountryLocalNamesNonConventional(System.Int64 key, System.Int64 relatedKey, [FromBody] Delta<CountryLocalNameUpdateDto> countryLocalName)
     {
         if (!ModelState.IsValid)
         {
@@ -137,7 +137,7 @@ public partial class CountriesController : ODataController
         return Ok(child);
     }
     
-    private async Task<CountryLocalNameDto?> TryGetCountryLocalName( System.Int64 key, CountryLocalNameKeyDto childKeyDto)
+    private async Task<CountryLocalNameDto?> TryGetCountryLocalName(System.Int64 key, CountryLocalNameKeyDto childKeyDto)
     {
         var parent = await _mediator.Send(new GetCountryByIdQuery(key));
         return parent?.CountryLocalNames.SingleOrDefault(x => x.Id == childKeyDto.keyId);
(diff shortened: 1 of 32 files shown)


rochar:
@anna-naumova25  take a look at this https://learn.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/odata-routing-conventions

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData.CountriesController.g.cs
```diff
@@ -91,9 +92,17 @@ public async Task<ActionResult> PostToCountryLocalNames([FromRoute] System.Int64
...
+        if (child == null)
+        {
+            return NotFound();
+        }
+        
+        return Created(child);
     }
     
+    [ODataIgnored]
+    [HttpPut("/api/Countries/{key}/CountryLocalNames/{relatedKey}")]
```
rochar, line 85: I tested this on our test app, I think we could do it like this, it will be available in Odata and Swagger:

```
  [HttpPut("api/Tenants/{key}/OwnerContacts/{relatedKey}")]        
  public async Task<ActionResult> PutToOwnerContactsNonOdataConvention([FromODataUri] int key, [FromODataUri] int relatedKey, [FromBody] OwnerContact contactPersonDetails)
  {
      return NotFound();
  }
```

image

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData.CountriesController.g.cs
```diff
@@ -107,9 +116,17 @@ public async Task<ActionResult> PutToCountryLocalNames([FromRoute] System.Int64
...
+        if (child == null)
+        {
+            return NotFound();
+        }
+        
+        return Ok(child);
     }
     
+    [ODataIgnored]
+    [HttpPatch("/api/Countries/{key}/CountryLocalNames/{relatedKey}")]
```
rochar, line 108: The same here

rochar (changes requested):
looks good, lets just align the endpoint convention

src/Nox.Generator/Presentation/Api/OData/EntityControllerGenerator.cs
```diff
@@ -112,10 +112,10 @@ public void Generate(SourceProductionContext context, NoxSolutionCodeGeneratorSt
             if (entity.Persistence is null ||
                 entity.Persistence.Read.IsEnabled)
```
anna-naumova25, line 113: @ricardo Is there any reason for generating Owned Relationships only if the Entity is readable?

rochar: this should appl;y only to the Get EndPoint....

rochar (approved):

PR #732, opened 2023-09-07
[Nox.Generators] Return 400 response when model is invalid #704

### LINKED ISSUE #704: [Nox.Generators] Return 400 response when model is invalid
See test "For now security check throws error, however response gets 500 instead of 400"

When a Model is invalid the response should be 400 bad request

AndreyDegtyarev:
Fixed StatudCode for bad request


rochar (approved):
nice job

PR #733, opened 2023-09-07
MP-9115 - Factory for converting Primitive types into Nox DateTimeRange

danield-ap:


tests/Nox.Lib.Tests/Factories/Types/NoxTypeDateTimeRangeFactoryTests.cs
```diff
@@ -0,0 +1,28 @@
...
+using Nox.Lib.Tests.FixtureConfig;
+using Nox.Solution;
+
+namespace Nox.Lib.Tests.Factories.Types;
+
+public record DateTimeRangeDto(DateTimeOffset Start, DateTimeOffset End);
+public class NoxTypeDateTimeRangeFactoryTests
+{
+    [Theory, AutoMoqData]
+    public void CreateNoxType_FromDto_IsValid(NoxSolution noxSolution, EntityDefinitionFixture fixture)
```
dejobratic, line 14: Can we add a test for null a value and an invalid value?

dejobratic (changes requested):
Looks good. We just need to add a few more tests.

tests/Nox.Lib.Tests/Factories/Types/NoxTypeDateTimeRangeFactoryTests.cs
danield-ap: Yes, tests added.

gjorgji-grgovski-authoritypartners (approved):

dejobratic (approved):

PR #734, opened 2023-09-07
Preparatory changes to solution for implementation of Integrations.

andresharpe:
- Fine-tuned solution definition and rules for Integrations
- Updated model and model validations
- Fixed all sample and test yaml solutions accordingly
- Fixed a couple of bugs found in validator


src/Nox.Solution/Models/Application/Integrations/Schedule/IntegrationSchedule.cs
```diff
@@ -20,5 +20,5 @@ public class IntegrationSchedule
 
     [Title("Check the ETL source for updates at appliation startup.")]
     [Description("Specify here whether the ETL source is checked for updates when the application starts.")]
-    public bool? RunOnStartup { get; internal set; }
+    public bool? RunOnStartup { get; internal set; } = false;
```
rochar, line 23: can we remove the nullable since is not required and the default is false? so we can avoid null checks in code?
```suggestion
    public bool RunOnStartup { get; internal set; } = false;
```

rochar (approved):
looks good, just one comment to review
and a test is failing

jan-schutte (approved):

jan-schutte (commented):
Just the tests that are failing

PR #735, opened 2023-09-07
MP-9122 - Factory for converting Primitive types into Nox type "Image"

danield-ap:


tests/Nox.Lib.Tests/Factories/Types/NoxTypeImageFactoryTests.cs
```diff
@@ -0,0 +1,27 @@
...
+using Nox.Lib.Tests.FixtureConfig;
+using Nox.Solution;
+
+namespace Nox.Lib.Tests.Factories.Types;
+
+public record ImageDto(string Url, string PrettyName, int SizeInBytes);
+public class NoxTypeImageFactoryTests
+{
+    [Theory, AutoMoqData]
+    public void CreateNoxType_FromDto_IsValid(NoxSolution noxSolution, EntityDefinitionFixture fixture)
```
dejobratic, line 13: Can we add a test for null a value and an invalid value?

dejobratic (changes requested):
Looks good. We just need to add a few more tests.

tests/Nox.Lib.Tests/Factories/Types/NoxTypeImageFactoryTests.cs
danield-ap: Yes, tests added.

gjorgji-grgovski-authoritypartners (approved):

dejobratic (approved):

rochar (approved):

PR #747, opened 2023-09-08
feature 742 Initial Commit Add Navigation Ui

### LINKED ISSUE #742: [Nox.Ui] Cryptocash.Ui add navigation menu ready to agree how generator creates content

DanielHodgeGH:
Initial flattened out NavigationMenu Links so that it is ready to be created by Nox.Generator - to be discussed

Also created IconHelper class for now to provide SVG paths for Icon whilst Entity yaml just has KeyEntity name for icon

Also moved Domain entities related Pages and Data items into Domain folders just to keep them separate


anna-naumova25 (approved):
Looks good, the only thing is to update entities from plural to singular (CustomersEntity -> CustomerEntity)

rochar (approved):
I have approved, so you and @anna-naumova25  can move forward with it when it makes sense

PR #758, opened 2023-09-11
MP-9136 - Change Formula type to return valid primitive type

danield-ap:


src/Nox.Types.Abstractions/Enums/Formula/FormulaReturnType.cs
```diff
@@ -7,12 +8,19 @@ namespace Nox.Types;
 /// </summary>
 public enum FormulaReturnType
 {
+    [Description("string")]
```
gjorgji-grgovski-authoritypartners: We don't need any data annotation here, just align  the Enum with the primitive types:
String -> string
Int -> int
....

src/Nox.Types.Abstractions/Extensions/NoxTypeExtensions.cs
```diff
@@ -74,4 +75,12 @@ public static bool IsUpdatableType(this NoxType noxType)
         return null;
     }
 
+    public static string ToPrimitiveType(this FormulaReturnType value)
```
gjorgji-grgovski-authoritypartners: We don't need this extension method.

src/Nox.Types/Types/Formula/Formula.cs
```diff
@@ -62,7 +63,7 @@ internal override ValidationResult Validate()
     /// Converts the instance of the object to string in format (returnType):expression.
     /// </summary>
     public override string ToString()
-        => $"({_formulaTypeOptions.Returns}):{Value}";
+        => $"({_formulaTypeOptions.Returns.ToPrimitiveType()}):{Value}";
```
gjorgji-grgovski-authoritypartners: No need for this change

gjorgji-grgovski-authoritypartners (changes requested):

gjorgji-grgovski-authoritypartners:
@danield-ap also rebuilt the Solution in order to re-generate all files and commit them as well.
Make sure that the tests are also passing.

src/Nox.Types/Types/Formula/Formula.cs
danield-ap: Change undone.

src/Nox.Types.Abstractions/Enums/Formula/FormulaReturnType.cs
danield-ap: `FormulaReturnType` enum options renamed.

src/Nox.Types.Abstractions/Extensions/NoxTypeExtensions.cs
danield-ap: Method removed.

danield-ap:
@gjorgji-grgovski-authoritypartners When I rebuild the whole solution I get 300+ files deleted and this error message `The analyzer assembly 'Nox.Generator.dll' references version '4.7.0.0' of the compiler, which is newer than the currently running version '4.4.0.0'`. 
I never get new files regenerated.

src/Nox.Types.Abstractions/Enums/Formula/FormulaReturnType.cs
```diff
@@ -7,12 +7,12 @@ namespace Nox.Types;
...
-    Double,
-    Decimal,
-    Bool,
+    @string,
+    @int,
+    @long,
+    @double,
+    @decimal,
+    @bool,
     DateTime,
```
gjorgji-grgovski-authoritypartners, line 16: Don't we need to have the @DateTime for the DateTime return type?

src/Nox.Solution/Extensions/ObjectExtensions.cs
```diff
@@ -266,4 +268,13 @@ private static string GetIndentation(int level)
     {
         return new string(' ', level * 4);
     }
+
+    private static bool IsReservedKeyword(string word)
```
gjorgji-grgovski-authoritypartners, line 272: Why do we need this?

gjorgji-grgovski-authoritypartners (commented):
Looks good, just small comment/question

src/Nox.Solution/Nox.Solution.csproj
```diff
@@ -24,6 +24,7 @@
 	<ItemGroup>
 		<PackageReference Include="FluentValidation" Version="11.7.1" />
 		<PackageReference Include="Humanizer.Core" Version="2.14.1" />
+		<PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.7.0" />
```
rochar, line 27: do we need this reference?

rochar (changes requested):

src/Nox.Types.Abstractions/Enums/Formula/FormulaReturnType.cs
danield-ap: No, the `@` is only needed for reserved keywords of the language, DateTime is just a class: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/?redirectedfrom=MSDN

src/Nox.Solution/Nox.Solution.csproj
danield-ap: Using this library was the easiest way to check for reserved keywords. I'm having to do this since we now have the FormulaReturnType enum using protected keywords and they were not being properly generated in the sample projects.
We are already using this library in other projects, so at least it's not a new library added to the solution.

src/Nox.Solution/Extensions/ObjectExtensions.cs
danield-ap: To fix the code generation of the Enum values with the reserved keywords.

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):

PR #763, opened 2023-09-12
MP-9129 - Improve the Soft Delete logic of the Auditable Entities with compound Nox Types

### LINKED ISSUE #725: [Nox.Lib] Improve the Soft Delete logic of the Auditable Entities with compound Nox Types [MP-9129]

dejobratic:
- Updated DbContext to reattach the deleted compound types when the parent entity is marked for deletion.
- Changed the store.entity.nox.yaml file:
    - Changed the id from nuid to databaseGuid, because the issue was observed on entities with databaseGuid as keys
    - Added multiple compound types, to make sure the reattaching logic works as expected
- Added tests for confirming that the soft delete works as expected


src/Nox.Generator/Infrastructure/Persistence/DbContext.template.cs
```diff
@@ -134,10 +135,25 @@ private void AuditEntity(EntityEntry<AuditableEntityBase> entry)
...
+                ReattachCompoundTypes(entry);
                 break;
         }
     }
 
+    private void ReattachCompoundTypes(EntityEntry<AuditableEntityBase> parentEntry)
+    {
+        foreach (var navigationEntry in parentEntry.Navigations)
+        {
+            foreach (var typeEntry in ChangeTracker.Entries<INoxType>())
```
rochar: @dejobratic  why INoxType and not IOwnedEntity? INoxType is for fields

src/Nox.ClientApi.Tests/.nox/design/Entitities/store.entity.nox.yaml
```diff
@@ -11,13 +11,8 @@ description: Stores
 keys:
 
   - name: Id
+    type: databaseGuid
     isRequired: true
-    description: NuidField Type
```
rochar, line 15: why the change, this way we lose the tests regarding Nuid?

src/Nox.ClientApi.Tests/Tests/Controllers/StoresControllerTests.cs
rochar, line 1: Can you also add test on Countries that does soft delete and thens tries to get a owned entity via /api/Countries/{key}/CountryLocalName/{relatedKey}, should return not found => ensure owned of soft deleted entities are not retrievable

rochar (changes requested):

src/Nox.Generator/Infrastructure/Persistence/DbContext.template.cs
dejobratic: The issue has been observed for Nox compound types, like StreetAddress and TranslatedText, which are registered in EF as owned entities, but they are of INoxType.

src/Nox.ClientApi.Tests/.nox/design/Entitities/store.entity.nox.yaml
dejobratic: We already have workplace.entity.nox.yaml with the same nuid key and configuration. 
image

And if I saw correctly, it has Nuid tests.
image

Therefore I decided to change the Store key to databaseGuid.

src/Nox.Generator/Infrastructure/Persistence/DbContext.template.cs
rochar: and other owned entities, for example CountryLocalNames in Countries, do we have the same issue?

src/Nox.ClientApi.Tests/.nox/design/Entitities/store.entity.nox.yaml
rochar: ok,

src/Nox.ClientApi.Tests/Tests/Controllers/StoresControllerTests.cs
dejobratic: Test case has been added. Thanks for the suggestion.

src/Nox.Generator/Infrastructure/Persistence/DbContext.template.cs
dejobratic: You are right. Those cases have been covered now.

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #769, opened 2023-09-12
MP-9144 remove databaseguid

msadikkoseaupa:


src/Nox.Generator/Application/Factories/EntityMapper.template.cs
```diff
@@ -30,7 +30,7 @@ public override void MapToEntity({{entity.Name}} entity, Entity entityDefinition
         dynamic? noxTypeValue;
     #pragma warning restore CS0168 // Variable is declared but never used
     {{ for key in entity.Keys -}}
-    {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "DatabaseGuid" || key.Type == "Formula" -}}
+    {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "Formula" -}}
```
msadikkoseaupa: Could you please pay extra attention to this and similar lines? @ggrgovski @dejobratic @rochar 
Some of them were removed completely whereas some of them were updated as Guid

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
```diff
@@ -20,7 +20,7 @@ namespace {{codeGeneratorState.ApplicationNameSpace }}.Dto;
 public partial class {{className}} : IEntityCreateDto<{{entity.Name}}>
 {
 {{- for key in entity.Keys }}
-    {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "DatabaseGuid" -}}
+    {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "Guid" -}}
```
gjorgji-grgovski-authoritypartners: We should remove this check if the type is "Guid" because we would like to have the possibility for setting up the Key by the user, even though we have set the default value in the factory.

gjorgji-grgovski-authoritypartners: We also need to make sure that for Guid type we do not have the Required attribute, as that one would be generated in the Factory if not provided.

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -53,11 +53,19 @@ public abstract class {{className}}Base: IEntityFactory<{{entity.Name}},{{entity
     {
         var entity = new {{codeGeneratorState.DomainNameSpace}}.{{entity.Name}}();
         {{- for key in entity.Keys }}
-            {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "DatabaseGuid" -}}
+            {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "Guid" -}}
```
gjorgji-grgovski-authoritypartners, line 56: We should remove this check if the type is "Guid"

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -53,11 +53,19 @@ public abstract class {{className}}Base: IEntityFactory<{{entity.Name}},{{entity
...
                 {{ continue; -}}
             {{- end }}
         entity.{{key.Name}} = {{ entity.Name }}.Create{{key.Name}}(createDto.{{key.Name}});
         {{- end }}
+        {{- for key in entity.Keys ~}}
+            {{- if key.Type == "Nuid" }}
+        entity.Ensure{{key.Name}}();
+            {{- end }}
+            {{- if key.Type == "Guid" }}
+        entity.{{key.Name}} = Nox.Types.Guid.From(System.Guid.NewGuid());
```
gjorgji-grgovski-authoritypartners: We should check if the CreateDto.{{key.Name}} is not provided, then to set new Guid for the entity.{{key.Name}}

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
gjorgji-grgovski-authoritypartners: This has already been changed if I'm not wrong in the latest version of the main branch.
However, Guid should work normally and call the Factory.

src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -46,6 +46,9 @@ public partial class {{className}}{{ if !entity.IsOwnedEntity }} : {{if entity.P
 		}
 	}
 
+	{{- else if key.Type == "Guid" -}}
```
gjorgji-grgovski-authoritypartners, line 73: I don't think you need to handle Guid type specifically, because it would generate the "else" section.
You would probably just need to add "Nox.Types" namespace before the "{{key.Type}}" on line 54.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
msadikkoseaupa: If we remove this check we will have two different lines for generating keyId for Guid types, so I belive this check should stay here. I am checking whether id is null or not and calling CreateId method from entity. 
What do you think ?

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
msadikkoseaupa: If we remove this line there will be double line for generating Id, so I belive this check should stay here. I check whether id is null or not and call Create method from entity.
What do you think?

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Generator/Application/Factories/EntityMapper.template.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
gjorgji-grgovski-authoritypartners: @msadikkoseaupa I'm not seeing the condition on the Required attribute

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
gjorgji-grgovski-authoritypartners: I don't think we need 2 "for" loops for the Keys, it would be much easier if we handled all the cases in only 1 (first one) "for" loop, but it is not wrong even like this.
@rochar I would wait for your feedback here, please.

src/Nox.Generator/Application/Commands/AddCommand.template.cs
rochar, line 1: Please  get the latest version from main... this should be done in a regular base and avoid having PR live for a long time

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
```diff
@@ -20,7 +20,7 @@ namespace {{codeGeneratorState.ApplicationNameSpace }}.Dto;
 public partial class {{className}} : IEntityCreateDto<{{entity.Name}}>
 {
 {{- for key in entity.Keys }}
-    {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "DatabaseGuid" -}}
+    {{- if key.Type == "Nuid" || key.Type == "AutoNumber" -}}
     {{ continue; -}}
     {{- end }}
     /// <summary>
```
rochar, line 36: @gjorgji-grgovski-authoritypartners  if the key is Guid, we have agreed that should not be required?

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
rochar: @msadikkoseaupa is correct on this part, but the ensureid is in the wrong place
Outdated:
is it not more simple a single for with if else?:
```
 {{- if key.Type == "Nuid" }}
        entity.Ensure{{key.Name}}();
{{- else if key.Type == "Guid" }}
         ....
{{- else if key.Type == "AutoNumber" }}
        {{ continue; -}}
{{ else }}
         entity.{{key.Name}} = {{ entity.Name }}.Create{{key.Name}}(createDto.{{key.Name}});
{{- end }}
```

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -53,11 +55,19 @@ public abstract class {{className}}Base: IEntityFactory<{{entity.Name}},{{entity
...
+            {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "Guid" -}}
                 {{ continue; -}}
             {{- end }}
         entity.{{key.Name}} = {{ entity.Name }}.Create{{key.Name}}(createDto.{{key.Name}});
         {{- end }}
+        {{- for key in entity.Keys ~}}
+            {{- if key.Type == "Nuid" }}
+        entity.Ensure{{key.Name}}();
+            {{- end }}
+            {{- if key.Type == "Guid" }}
```
rochar, line 67: use else and avoid cascading if

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Factories.StoreOwnerFactory.g.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Generated
+﻿using System;// Generated
```
rochar: generated should be in a single line in a top

tests/Nox.Types.Tests/EntityFrameworkTests/Configuration/CountryDbConfiguration.cs
```diff
@@ -50,7 +50,7 @@ public void Configure(EntityTypeBuilder<Country> builder)
         builder.Property(e => e.InfoEmail).HasConversion<EmailConverter>();
         builder.Property(e => e.SecretPassword).HasConversion<EncryptedTextConverter>();
         builder.Property(e => e.AutoId).ValueGeneratedOnAdd().HasConversion<AutoNumberConverter>();
-        builder.Property(e => e.DatabaseGuid).ValueGeneratedOnAdd().HasConversion<DatabaseGuidConverter>();
+        builder.Property(e => e.Guid).ValueGeneratedOnAdd().HasConversion<GuidConverter>();
```
rochar: do we want ValueGeneratedOnAdd?

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Factories.WorkplaceFactory.g.cs
```diff
@@ -39,8 +39,8 @@ public virtual Workplace CreateEntity(WorkplaceCreateDto createDto)
     private ClientApi.Domain.Workplace ToEntity(WorkplaceCreateDto createDto)
     {
         var entity = new ClientApi.Domain.Workplace();
+        entity.EnsureId();
```
rochar: EnsureId needs to be the last method to be call because we will need the entity to be properly set

Strange that there is not test failing because of this change! Can you please confirm?

rochar (changes requested):

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Factories.WorkplaceFactory.g.cs
rochar: run a manual build and confirm is green before having the PR

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -53,11 +55,19 @@ public abstract class {{className}}Base: IEntityFactory<{{entity.Name}},{{entity
     {
         var entity = new {{codeGeneratorState.DomainNameSpace}}.{{entity.Name}}();
         {{- for key in entity.Keys }}
-            {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "DatabaseGuid" -}}
+            {{- if key.Type == "Nuid" || key.Type == "AutoNumber" || key.Type == "Guid" -}}
                 {{ continue; -}}
             {{- end }}
         entity.{{key.Name}} = {{ entity.Name }}.Create{{key.Name}}(createDto.{{key.Name}});
         {{- end }}
+        {{- for key in entity.Keys ~}}
```
rochar: The Nuid part should be in the bottom after all fields are assigned... a function may use one or more entity fields

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -53,11 +55,19 @@ public abstract class {{className}}Base: IEntityFactory<{{entity.Name}},{{entity
...
+            {{- if key.Type == "Nuid" }}
+        entity.Ensure{{key.Name}}();
+            {{- end }}
+            {{- if key.Type == "Guid" }}
+        entity.{{key.Name}} = ( createDto.{{key.Name}} == null || System.Guid.Empty.Equals(createDto.{{key.Name}}) ) ? Nox.Types.Guid.From(System.Guid.NewGuid()) : {{ entity.Name }}.Create{{key.Name}}(createDto.{{key.Name}});
```
rochar: We should call a virtual function on the Entity, so it can be overridden, like EnsureId, you will need to update the Entity.Template

`entity.{{key.Name}} =  {{ entity.Name }}.Ensure{{key.Name}}(createDto.{{key.Name}});`

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
rochar: the Guid part can/should be moved to the first loop

tests/Nox.Types.Tests/EntityFrameworkTests/Configuration/CountryDbConfiguration.cs
msadikkoseaupa: Actually no, fixed

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Factories.WorkplaceFactory.g.cs
msadikkoseaupa: Moved to the end of method

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Factories.StoreOwnerFactory.g.cs
msadikkoseaupa: Using statement was redundant based on the latest implementation and it was removed

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Generator/Application/Commands/AddCommand.template.cs
msadikkoseaupa: Branch was updated

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
msadikkoseaupa: Since Nuid needs to be end of the method, it stays as is. 
Could you please check the latest implementation.

rochar (dismissed):

gjorgji-grgovski-authoritypartners (dismissed):

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #785, opened 2023-09-13
Feature/772 noxgenerators create endpoints for single owned entity without relatedkey like multiple

### LINKED ISSUE #772: [Nox.Generators] Create endpoints for single Owned Entity (without relatedKey) like multiple

anna-naumova25:


src/Nox.Generator/Application/Commands/AddCommandGenerator.cs
```diff
@@ -23,11 +24,12 @@ public void Generate(SourceProductionContext context, NoxSolutionCodeGeneratorSt
         {
             context.CancellationToken.ThrowIfCancellationRequested();
 
-            var parent = codeGeneratorState.Solution.Domain.Entities.FirstOrDefault(e =>
-                e.OwnedRelationships?.Any(o => o.Entity == entity.Name && !o.WithSingleEntity) == true);
+            var parent = entity.TryGetParent(codeGeneratorState.Solution.Domain.Entities);
             if (parent is null)
                 continue;
 
+            var isSingleRelationship = parent.OwnedRelationships.First(o => o.Entity == entity.Name).WithSingleEntity;
```
rochar, line 31: this is tricky because we are assuming an entity can only be owned by another and not by two for example

we need to prepare the code to generate a command for each parent. lets discuss this

rochar (approved):
we just need to discuss and create a ticket to support a entity to be owned by different ones

PR #786, opened 2023-09-13
MP-9223 Generate Domain Events (records) for the EntityGenerator

msadikkoseaupa:


src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -5,11 +5,33 @@
...
 using System.Collections.Generic;
 
-using Nox.Types;
+using Nox.Abstractions;
 using Nox.Domain;
+using Nox.Types;
 
 namespace {{codeGeneratorState.DomainNameSpace}};
 
+/// <summary>
```
dejobratic, line 18: The comments should be withing the if statement as well. Otherwise, we will have comments without domain event records where *.RaiseEvents == false.

dejobratic (approved):
Looks good.

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -5,11 +5,33 @@
...
 using System.Collections.Generic;
 
-using Nox.Types;
+using Nox.Abstractions;
 using Nox.Domain;
+using Nox.Types;
 
 namespace {{codeGeneratorState.DomainNameSpace}};
 
+{{- if entity.Persistence.Create.RaiseEvents == true}}
```
rochar:
```suggestion
{{- if entity.Persistence.Create.RaiseEvents}}
```

rochar (approved):
looks good, just a simple change proposal

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: Desired changes were applied

PR #793, opened 2023-09-14
570 Create a new Entity and related entity in the same Post

### LINKED ISSUE #570: [Nox.Generators] Create a new Entity and related entity in the same Post
relates #776 
### Create a new Entity and related entity in the same Post
1. Post allows to create and associate a related entity via navigation property 
   1. CreatedDto.template changes 
        2. Remove the Id Property for single entity
        3. Add the navigation property to <RelatedEntity>CreateDto with the name of the Relation
        Example:
        ```
        public virtual StoreOwnerCreateDto? Ownership { get; set; } = null!;
        ```
   1. Update factory to create related entity (using the related entity factory)
   5. Create a virtual method CreateRefTo\<EntityName\>(relatedKey) in the entity. Call it from the factory
1.  (**Pending**) If relation is one* and the related entity already exists how to associate it? Need an example where this makes sense

Example:

```
public abstract class StoreFactoryBase: IEntityFactory<Store,StoreCreateDto>
{

 private ClientApi.Domain.Store ToEntity(StoreCreateDto createDto)
 {
     var entity = new ClientApi.Domain.Store();
     entity.Name = ClientApi.Domain.Store.CreateName(createDto.Name);
		entity.EnsureId();
     var store = StoreOwnerFactory.CreateEntity(createDto.StoreOwner);
     entity.CreateRefToStoreOwner(store);
     return entity;
 }
...
}

public partial class Store : AuditableEntityBase, IEntityConcurrent
{
    public virtual void CreateRefToStoreOwner(StoreOwner storeOwner)
    {
        Ownership = storeOwner;
        //override business logic
    })
...
}
```

anna-naumova25:
I will use Country to Workplace relation to generate tests as it is done for Owned relationships.
It is ZeroOrMany on Country side and ZeroOrOne on Workplace side.


src/Nox.Generator/Application/Queries/ByIdQuery.template.cs
anna-naumova25, line 1: @rochar Please see if you're ok with this change. I added it so EF includes related entities. Then we can use OData expand to receive the related entity/list.

rochar:
Can you also add an example where one of the side is exactlyOne or OneOrMany? so we can see how it behaves

thks

src/Nox.Generator/Application/Queries/ByIdQuery.template.cs
rochar: I think OData Expand does this on IQueryable automatically, we should not need to do it explicitly...., its like paging,,, on the IQueryable Rrsult it adds additional filter/includes/etc... can you double check?

The trade-off is that we are getting a lot info from the db, even if its not needed

rochar (approved):
Look good, we just need to check the expand

PR #804, opened 2023-09-18
MP-9270 - OwnedEntity Validation - Can not have Relationships

### LINKED ISSUE #792: [Nox.Solution] OwnedEntity Validation - Can not have Relationships MP-9270
Add a new validation to Nox.Solution:

An OwnedEntity (exists in an second Entity  OwnedRelationships) can not have have a relationship to a second Entity (cannot have Relationships), because the Owned Entity it will not be treated as an Entity

dejobratic:
- An owned entity cannot have relationships to other entities
- An owned entity cannot be related to other entities (cannot be referenced in other entities relationships)


src/Nox.Solution/Validation/Domain/OwnedEntityValidator.cs
```diff
@@ -0,0 +1,51 @@
...
+            });
+        });
+
+        // Owned entity cannot be auditable
+        RuleFor(e => e.Persistence!.IsAudited)
+            .NotEqual(true)
+            .WithMessage(e => string.Format(ValidationResources.EntityOwnedCannotBeAuditable, e.Name));
+
+        // Owned entity cannot have relationships to other entities
+        RuleFor(e => e.Relationships.Any())
```
gjorgji-grgovski-authoritypartners, line 39: FluentValidation evaluates the rules in the order they are defined and stops as soon as it encounters a validation failure, so please move this validation rule to the first one in the OwnedEntityValidator.

src/Nox.Solution/Validation/Domain/OwnedEntityValidator.cs
```diff
@@ -0,0 +1,51 @@
...
+            .NotEqual(true)
+            .WithMessage(e => string.Format(ValidationResources.EntityOwnedCannotHaveRelationships, e.Name, e.OwnerEntity?.Name));
```
gjorgji-grgovski-authoritypartners: You have only 1 placeholder in the Resource phrase, so it would be replaced by the "e.Name" and "e.OwnerEntity?.Name" is redundant in this case.

src/Nox.Solution/Validation/Domain/OwnedEntityValidator.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        // Owned entity cannot be related to other entities
+        RuleFor(e => IsRelatedToOtherEntities(e, entities))
+            .NotEqual(true)
+            .WithMessage(e => string.Format(ValidationResources.EntityOwnedCannotBeRelatedToOtherEntities, e.Name, e.OwnerEntity?.Name));
```
gjorgji-grgovski-authoritypartners: You have only 1 placeholder in the Resource phrase, so it would be replaced by the "e.Name" and "e.OwnerEntity?.Name" is redundant in this case.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Solution/Validation/Domain/OwnedEntityValidator.cs
dejobratic: The default behavior of the validator is to run all of the rules. If we want the validator to stop on first failure, then we would need to specify the validator-level casccade mode. https://docs.fluentvalidation.net/en/latest/cascade.html

src/Nox.Solution/Validation/Domain/OwnedEntityValidator.cs
dejobratic: Changed. Thanks.

src/Nox.Solution/Validation/Domain/OwnedEntityValidator.cs
dejobratic: Changed. Thanks.

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #818, opened 2023-09-20
Feature/#774 refactor entity mapper map to entity

### LINKED ISSUE #774: [Nox.Lib] Refactor EntityMapper MapToEntity
Goal: Remove MapToEntity from IEntityMapper and avoid dynamic types

1. Rename the  IEntityCreateDto to IEntityDto, no need to distinguish all Dto
1. Update the IEntityFactory, adding a new method UpdatedEntity

```
 /// <summary>
 /// Factory for Entities
 /// </summary>
 public interface IEntityFactory<E,D> where E : IEntity where D : IEntityDto<E> 
 {
     /// <summary>
     /// Create an entity from a Dto instance
     /// </summary>
     /// <param name="createDto"></param>
     /// <returns></returns>
     E CreateEntity(D createDto);

     /// <summary>
     /// Updates an entity using a dto
     /// </summary>
     /// <param name="entity">Entity to Update</param>
     /// <param name="updateDto">Dto</param>
     void UpdatedEntity(E entity, D updateDto);
 }
```

1. Generated the UpdateEntity in the same way we are generating the CreateEntity, but receiving the entity => We may create a private method to update the entity and call it in the public methods
    1. Do Not Include the Owned Entities full update  this is down via the owned entities endpoints to avoid complexity:
    
image

3. remove the MapToEntity from IEntityMapper
4. refactor the code that uses MapToEntity to use IEntityFactory. UpdatedEntity

ILoveMyTortoise:
closes #774


rochar (dismissed):
good job

PR #831, opened 2023-09-21
MP-9108 - NOX - Refactor Auditable Properties to be Primitive Types

emir-sehovic1:
- Refactored audit properties to be primitive types
- Regenerated code
- Removed previous auditable fields db configurator


samples/Cryptocash.Api/Program.cs
```diff
@@ -1,10 +1,18 @@
...
 
+using System.Text.Json.Serialization;
+
 var builder = WebApplication.CreateBuilder(args);
 
 // Add services to the container.
 
-builder.Services.AddControllers();
+builder.Services.AddControllers()
+    .AddJsonOptions(options =>
```
gjorgji-grgovski-authoritypartners: Why do we need this JsonSerializerOptions?

gjorgji-grgovski-authoritypartners:
@emir-sehovic1 please add the MP number as well in the title

samples/Cryptocash.Api/Program.cs
emir-sehovic1: It is not strictly related to this feature, but instead this is more of a general change because Systm.Text.Json serializer is case sensitive and also doesn't deserialize enum strings to enum values by default. So with this it is easier to test and consume cryptocash api.

gjorgji-grgovski-authoritypartners: @rochar are you fine with this having it on the Service level, or we would introduce it to the generated Extension maybe for the generated Controllers?
By the way... this is a generic setting also fo the newly added controllers, which don't have to be OData Controllers.

rochar: @emir-sehovic1 in that case I would prefer to remove it, if this is a service configuration let the person that is doing CryptoCash to define if its needed. You can speak with @DanielHodgeGH  about it if y want.

rochar (dismissed):

samples/Cryptocash.Api/Program.cs
emir-sehovic1: I've removed it from this PR, will check with Daniel to add it if he needs it. Thank you.

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #852, opened 2023-09-25
MP-9230 domain default events

msadikkoseaupa:


src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -18,27 +19,27 @@ public partial class {{className}}:{{className}}Base
 /// <summary>
 /// Record for {{entity.Name}} created event.
 /// </summary>
-public record {{entity.Name}}Created({{entity.Name}} {{entity.Name}}) : IDomainEvent;
+public record {{entity.Name}}Created({{entity.Name}}Base {{entity.Name}}) : IDomainEvent;
```
rochar: Is there a reason to add the Base class here?

tests/Nox.Generator.Tests/Infrastructure/Persistence/DatabaseServerTests.cs
```diff
@@ -49,11 +50,16 @@ public void Can_generate_database_server_files()
         var countryFileName = "Domain.Country.g.cs";
         Assert.True(generatedSources.Any(s => s.HintName == countryFileName), $"{countryFileName} not generated");
         var generated = generatedSources.First(s => s.HintName == countryFileName).SourceText.ToString();
-        Assert.Equal(File.ReadAllText("./ExpectedGeneratedFiles/Country.expected.g.cs"), generated);
+        Assert.Equal(Normalize(File.ReadAllText("./ExpectedGeneratedFiles/Country.expected.g.cs")), Normalize(generated));
```
rochar: why normalize this one only? we compare what we expected to what is being generated...

rochar (changes requested):

src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -108,6 +109,46 @@ public abstract class {{className}}Base{{ if !entity.IsOwnedEntity }} : {{if ent
...
+	{{- if entity.Persistence.Create.RaiseEvents }}
+	public virtual void RaiseCreateEvent()
+	{
+		_domainEvents.Add(new {{entity.Name}}Created(this));     
+	}
+	{{ else }}
+	public virtual void RaiseCreateEvent(){ }
+	{{- end }}
+	///<inheritdoc/>
+	{{- if entity.Persistence.Update.RaiseEvents }}
```
gjorgji-grgovski-authoritypartners: I think it would be much more clear if we made this check inside the implementation and would not need the "else" block

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: Yes, since we add raise domain event methods inside base class, "this" keyword means base class. 

_domainEvents.Add(new EntityCreated(this);

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: My initial implementation was like your mention. On the other hand I would prefer single line method having openning and closing brackets if it is empty. 
What do you think?

src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -108,6 +109,46 @@ public abstract class {{className}}Base{{ if !entity.IsOwnedEntity }} : {{if ent
     public Nox.Types.{{attribute.Type}}{{if !attribute.IsRequired}}?{{end}} {{attribute.Name}} { get; set; } = null!;
     {{- end}}
 {{- end }}
+{{-if entity.HasDomainEvents}}
+
+	///<inheritdoc/>
+	public IReadOnlyCollection<IDomainEvent> DomainEvents => _domainEvents;
+
+	private readonly List<IDomainEvent> _domainEvents = new();
```
rochar: can you make this protected, so each service on the partial class can add or override methods and acess the _domainEvents

src/Nox.Generator/Domain/Entity.template.cs
gjorgji-grgovski-authoritypartners: That is just formatting, so no need to worry about that in this case.
Let's keep the template a bit simpler.

src/Nox.Generator/Domain/Entity.template.cs
rochar: The event needs to be raised with the actual entity that was created for example, so I would suggest something like this:

image

this way a service implementation, like crypto cash,  can override the raiseinternal and extend the behavior, that is the reason why we have a base class with all the implementation

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: desired changes were applied

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: Desired changes were applied

tests/Nox.Generator.Tests/Infrastructure/Persistence/DatabaseServerTests.cs
msadikkoseaupa: Actually I have trouble with line endings since dev environemnt is unix based. I found another workaround and removed the normalize method.

src/Nox.Generator/Domain/Entity.template.cs
msadikkoseaupa: I updated the implementation, this time I need to move IEntityHaveDomainEvents interface to actual entity.

src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -87,6 +87,12 @@ public class Entity : DefinitionBase
 
     [YamlIgnore]
     public Entity? OwnerEntity { get; internal set; }
+
```
rochar, line 90: @dejobratic you will have to adjust on your PR use the new function created ..

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #855, opened 2023-09-26
Compile tests added for generation tests

msadikkoseaupa:


src/Nox.Generator/Application/Commands/CreateCommandGenerator.cs
```diff
@@ -25,10 +25,14 @@ public void Generate(SourceProductionContext context, NoxSolutionCodeGeneratorSt
 
             var primaryKeysQuery = string.Join(", ", entity.Keys.Select(k => $"entityToCreate.{k.Name}.Value"));
 
+            //TestEntityTwoRelationshipsOneToOne has more than one relationship to SecondTestEntityTwoRelationshipsOneToOne
```
rochar: we do not want this comments on the template?

src/Nox.Generator/Application/Commands/CreateCommand.template.cs
```diff
@@ -1,4 +1,3 @@
-﻿{{- relatedEntities = entity.Relationships | array.map "Entity" }}
```
rochar, line 1: so were do we compute the related entities?

src/Nox.Generator/Application/Commands/CreateCommand.template.cs
rochar, line 1: add also the new generated code, so we can see the differences

tests/Nox.Integration.Tests/CompileTests.cs
rochar, line 1: Can we move this test to Nox.Generator.Tests?

rochar (changes requested):

tests/Nox.Integration.Tests/CompileTests.cs
msadikkoseaupa: If we copy exactly same nox.yaml files with this project it is possible. Because this project covers many edge cases.

src/Nox.Generator/Application/Commands/CreateCommandGenerator.cs
msadikkoseaupa: This a comment for your attention actually, we calculate the related entities here with group by method.

src/Nox.Generator/Application/Commands/CreateCommand.template.cs
msadikkoseaupa: we compute related entites in CreateCommandGenerator.cs lines between 28 to 33

rochar (dismissed):

rochar (approved):

dejobratic (approved):

PR #859, opened 2023-09-26
Feature/Store type flagged db providers

jan-schutte:


src/Nox.Types.EntityFramework/Enums/NoxDataStoreTypeFlags.cs
```diff
@@ -0,0 +1,10 @@
+namespace Nox.Types.EntityFramework.Enums;
+
+[Flags]
+public enum NoxDataStoreTypeFlags
+{
+    EntityStore,        //Persists domain entities as defined in the solution.nox.yaml                             --
```
rochar: you could add the comment as Summary... so devs will see it more easily

src/Nox.EntityFramework.Postgres/PostgresDatabaseProvider.cs
```diff
@@ -4,18 +4,20 @@
...
 using Nox.Types.EntityFramework.Configurations;
 using Nox.Types.EntityFramework.EntityBuilderAdapter;
+using Nox.Types.EntityFramework.Enums;
 using Npgsql;
 
 namespace Nox.EntityFramework.Postgres;
 
 public class PostgresDatabaseProvider: NoxDatabaseConfigurator, INoxDatabaseProvider 
 {
+    public string ConnectionString { get; protected set; } = string.Empty;
```
rochar, line 14: if this is for all providers? you can add them to the base class NoxDatabaseConfigurator

src/Nox.EntityFramework.Postgres/PostgresDatabaseProvider.cs
```diff
@@ -54,4 +56,14 @@ public string ToTableNameForSqlRaw(string table, string schema)
     {
         throw new NotImplementedException();
     }
+
```
rochar, line 59: if this is for all providers? you can add them to the base class NoxDatabaseConfigurator

src/Nox.Types.EntityFramework/Enums/NoxDataStoreTypeFlags.cs
jan-schutte: added to summary

src/Nox.EntityFramework.Postgres/PostgresDatabaseProvider.cs
jan-schutte: Rather keep it on the provider, not all database provider will implement the configurator

src/Nox.EntityFramework.Postgres/PostgresDatabaseProvider.cs
jan-schutte: Rather keep it on the provider, not all database provider will implement the configurator

rochar (approved):

PR #864, opened 2023-09-26
Feature/836 noxgenerator add validate to entitydto

### LINKED ISSUE #836: [Nox.Generator] Add Validate to <Entity>Dto
Update the \<Entity\>Dto so it can be used on the UI to validating user input, leveraging our Nox.Types.

1. Create a Validate method
2. Iterate all Entity Properties
3. Instantiate the Nox.Type for each property using the current  DTO value
4. Collect all Validation Errors and return `IReadonlyDictionary<string,Nox.Types.ValidationResult>` per field 
5. Consider required vs not required fields

pseudo code
```
public virtual IReadOnlyDictionary<string, IEnumerable<ValidationResult>> Validate()
{
    var result;
    try
    {
        Nox.Type.Text CreateName(this.Name)
    }
    catch (TypeValidationException e)
    {
        ///result[e.Variable] = e.ValidationResults;         
    }
}
```

anna-naumova25:
@rochar there are 3 points to think about:

- `private void ValidateField<T>(string fieldName, Func<T> action, Dictionary<string, IEnumerable<string>> result)`
this function can be moved to a separate file so we generate it once, not in each class. @rochar could you please suggest the best class to move this function to? 
- do we need to add any validation for Keys?
- we can wrap everything inside Validate function into the additional try/catch block to avoid unexpected exceptions during the Validation


src/Nox.Generator/Application/Dto/EntityDto.template.cs
```diff
@@ -22,6 +22,48 @@ public record {{entity.Name}}KeyDto({{primaryKeys}});
...
+        {{- else }}
+            ValidateField("{{attribute.Name}}", () => {{codeGeneratorState.DomainNameSpace}}.{{entity.Name}}.Create{{attribute.Name}}(this.{{attribute.Name}}.NonNullValue<{{attribute.Type}}Dto>()), result);
+        {{- end}}
+        {{- end }}
+    {{- end}}
+
+        return result;
+    }
+
+    private void ValidateField<T>(string fieldName, Func<T> action, Dictionary<string, IEnumerable<string>> result)
```
rochar: maybe we can create a` EntityDtoBase<T>: IEntityDto<T>` and add this method

in Nox.Lib.Application.Dto?

tests/Nox.Generator.Tests/ExpectedGeneratedFiles/Application.Dto.CountryDto.expected.g.cs
```diff
@@ -23,6 +23,52 @@ public record CountryKeyDto(System.String keyId);
...
+    {
+        try
+        {
+            action();
+        }
+        catch (TypeValidationException ex)
+        {
+            result.Add(fieldName, ex.Errors.Select(x => x.ErrorMessage));
+        }
+        catch (NullReferenceException)
```
rochar: not sure if this will work for all noxTypes

may be we should be explicit in the Validate, for example add for all required

 if (this.Name is  null)
{
      result.Add("Name", new [] { $"{fieldName} is Required." });
}
else
{
       ValidateField("Name", () => SampleWebApp.Domain.Country.CreateName(this.Name), result);
}

what o you think?

tests/Nox.Generator.Tests/ExpectedGeneratedFiles/Application.Dto.CountryDto.expected.g.cs
```diff
@@ -23,6 +23,52 @@ public record CountryKeyDto(System.String keyId);
...
+        {
+            result.Add(fieldName, new List<string> { $"{fieldName} is Required." });
```
rochar: for cases like this where you know the size of collection prefer to use an array, and improve performance
```suggestion
            result.Add(fieldName, new [] { $"{fieldName} is Required." });
```

rochar (changes requested):
lets talk about the comments
thanks

rochar:
> @rochar there are 3 points to think about:
> 
> * `private void ValidateField<T>(string fieldName, Func<T> action, Dictionary<string, IEnumerable<string>> result)`
>   this function can be moved to a separate file so we generate it once, not in each class. @rochar could you please suggest the best class to move this function to?
> * do we need to add any validation for Keys?
> * we can wrap everything inside Validate function into the additional try/catch block to avoid unexpected exceptions during the Validation

regarding the function I left a suggestion on my comments

Keys, well if we assume they are going to be:

1. Nuid => No Validation
2. Guid => No Validation
3. AutoNumber => No Validation
4. Entity ForeignKey => Required => And its Guid/Nui/AutoNumber => We need to validate is set on this case

on the try catch jus catch expected exceptions that we now how to handle, for now I think this is what we want

DanielHodgeGH:
Thanks @anna-naumova25 - looks good - I cant add any more than what @rochar said - please request a re-review once those changes are made thanks Dan

tests/Nox.Generator.Tests/ExpectedGeneratedFiles/Application.Dto.CountryDto.expected.g.cs
anna-naumova25: if(typeof(this.Name).IsValueType)
{
ValidateField("Name", () => SampleWebApp.Domain.Country.CreateName(this.Name), result);
}
else 
{
result.Add("Name", new [] { $"{fieldName} is Required." });
}

src/Nox.Generator/Application/Dto/EntityDto.template.cs
anna-naumova25: rename the method and the params

DanielHodgeGH (dismissed):
Thanks Anna looks good

src/Nox.ClientApi.Tests/Tests/Application/Dto/CountryQualityOfLifeIndexDtoValidatorTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            //Act
+            var validationResult = indexDto.Validate();
+
+            //Assert
+            validationResult.Should().NotBeNull();
+            validationResult.Should().ContainKey(nameof(indexDto.CountryId));
+            validationResult.Should().ContainKey(nameof(indexDto.IndexRating));
+        }
+
+        [Fact(Skip = "System.InvalidOperationException : AutoNumber can only be created with FromDatabase.")]
```
anna-naumova25, line 50: @rochar @ILoveMyTortoise Skip should be removed once the Key Relationships are fixes

src/Nox.Lib/Application/Dto/EntityDtoBase.cs
```diff
@@ -0,0 +1,22 @@
...
+using Nox.Types;
+
+namespace Nox.Application.Dto;
+
+/// <summary>
+/// Base Implementation for a Nox EntityDto
+/// </summary>
+public abstract class EntityDtoBase
+{
+    public void TryGetValidationExceptions<T>(string propertyName, Func<T> createAction, Dictionary<string, IEnumerable<string>> validationResult)
```
anna-naumova25: @rochar so for now it's called TryGetValidationExceptions. Please see my other options, let me know if you want to change it:

- CreateAndValidate
- CreateWithValidation
- TryGetValidationExceptions
- TryGetValidationExceptionsOnCreate

rochar: The TryGetValidationExceptions leads to expecting a bool result and a list of exception, I could suggest 

ExecuteActionAndCollectValidationExceptions , too much verbose but I would prefer this one :)

rochar (dismissed):

PR #868, opened 2023-09-27
834-2 Add UserInterface to EntityMetadata

### LINKED ISSUE #834: [Nox.Generator] Update Entity Metadata to include UI settings
We are generating Metada extensions for entities to keep the default options and be able to create Nox.type with does options:

Image

In the scope of this ticket:

1. (PR1)Remove Domain from the file path generated
2. (PR1)Rename the class to \<Entity\>Metadata and make the \<Entity\> internal (so it can not be accessible outside the domain)
3. (PR2)Update the generator and include the UI settings per property
    4. (PR2)Use the template scriban as usual, do not try to use the current structure this is not the path that we want to take

Image

Consider something like this

`public static TypeUserInterface OpeningDayUSerInterface(NoxSolution solution) => solution.Domain.Entities["Country"].Attributes["OpeningDay"].UserInterface;`

anna-naumova25:
@DanielHodgeGH you'll need to inject NoxSolution via the constructor into the class where you want to access UserInterfaces. 

@rochar I cancelled the automatic build as it took 2+ hours. Will take a look at it tomorrow to see if it was caused by my changes or anything else in main.

### DIFF
--- a/samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.BankNoteMetadata.g.cs
+++ b/samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.BankNoteMetadata.g.cs
@@ -4,6 +4,7 @@
 
 using Nox.Types;
 using Nox.Domain;
+using Nox.Solution;
 using System;
 using System.Collections.Generic;
 
@@ -48,4 +49,22 @@ public partial class BankNoteMetadata
         public static Nox.Types.Money CreateValue(IMoney value)
             => Nox.Types.Money.From(value);
         
+
+        /// <summary>
+        /// User Interface for property 'CashNote'
+        /// </summary>
+        public TypeUserInterface? CashNoteUserInterface(NoxSolution solution) 
+            => solution.Domain?
+                .Entities?.FirstOrDefault(e => e.Name == "BankNote")?
+                .Attributes?.FirstOrDefault(a => a.Name == "CashNote")?
+                .UserInterface;
+
+        /// <summary>
+        /// User Interface for property 'Value'
+        /// </summary>
+        public TypeUserInterface? ValueUserInterface(NoxSolution solution) 
+            => solution.Domain?
+                .Entities?.FirstOrDefault(e => e.Name == "BankNote")?
+                .Attributes?.FirstOrDefault(a => a.Name == "Value")?
+                .UserInterface;
 }
--- a/samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.BankNoteMetadata.g.cs
+++ b/samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.BankNoteMetadata.g.cs
@@ -53,18 +53,18 @@ public partial class BankNoteMetadata
         /// <summary>
         /// User Interface for property 'CashNote'
         /// </summary>
-        public TypeUserInterface? CashNoteUserInterface(NoxSolution solution) 
-            => solution.Domain?
-                .Entities?.FirstOrDefault(e => e.Name == "BankNote")?
-                .Attributes?.FirstOrDefault(a => a.Name == "CashNote")?
+        public static TypeUserInterface? CashNoteUserInterface(NoxSolution solution) 
+            => solution.Domain!
+                .GetEntityByName("BankNote")
+                .GetAttributeByName("CashNote")?
                 .UserInterface;
 
         /// <summary>
         /// User Interface for property 'Value'
         /// </summary>
-        public TypeUserInterface? ValueUserInterface(NoxSolution solution) 
-            => solution.Domain?
-                .Entities?.FirstOrDefault(e => e.Name == "BankNote")?
-                .Attributes?.FirstOrDefault(a => a.Name == "Value")?
+        public static TypeUserInterface? ValueUserInterface(NoxSolution solution) 
+            => solution.Domain!
+                .GetEntityByName("BankNote")
+                .GetAttributeByName("Value")?
                 .UserInterface;
 }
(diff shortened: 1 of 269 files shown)


samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.VendingMachineMetadata.g.cs
DanielHodgeGH, line 1: Looks good - but more of a question - so we have typeoptions and userinterface attributes now accessible through meta - where does the yaml for properrty 'isRequired' sit? - or shall I deduce that on the UI side from the fact the DTO property is not nullable ?

Also im assuming in reality it will be the generator accessing the userinteface properties rather than UI app itself injecting NoxSolution - for example so for text input with UI.inputmask - the generator will produce the blazor code for a textbox with input mask already setup - we can obviously discuss when it comes to integrating generator itself

DanielHodgeGH (dismissed):

samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.VendingMachineMetadata.g.cs
anna-naumova25: isRequired is on the same level as typeoptions and user interface. Please see below, it is taken from Cryptocash:
```
  - name: Name
    description: Country's name
    type: text
    textTypeOptions:
      minLength: 4
      maxLength: 63
    isRequired: true
    userInterface:
      showInSearchResults: optionalAndOnByDefault
      canSort: true
      canSearch: true
      canFilter: true
      showOnCreateForm: true
      showOnUpdateForm: true
```

For the second option - yes, I totally agree that it's better to use userInterface params on the generation phase not from the UI app.

src/Nox.Generator/Domain/EntityMeta.template.cs
```diff
@@ -40,4 +41,16 @@ public partial class {{className}}
...
     {{ end }}
 {{- end }}
+
+    {{- for attribute in entity.Attributes }}
+
+        /// <summary>
+        /// User Interface for property '{{attribute.Name}}'
+        /// </summary>
+        public static TypeUserInterface? {{attribute.Name}}UserInterface(NoxSolution solution) 
+            => solution.Domain?
```
rochar: the null checks are unnecessary otherwise the property would not be generated

src/Nox.Generator/Domain/EntityMeta.template.cs
```diff
@@ -40,4 +41,16 @@ public partial class {{className}}
...
+                .Entities?.FirstOrDefault(e => e.Name == "{{entity.Name}}")?
```
rochar: use GetEntityByName will be more efficient

src/Nox.Generator/Domain/EntityMeta.template.cs
```diff
@@ -40,4 +41,16 @@ public partial class {{className}}
...
+                .Attributes?.FirstOrDefault(a => a.Name == "{{attribute.Name}}")?
```
rochar: use GetAttributeByName also

samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.BankNoteMetadata.g.cs
```diff
@@ -48,4 +49,22 @@ public static Nox.Types.Text CreateCashNote(System.String value)
         public static Nox.Types.Money CreateValue(IMoney value)
             => Nox.Types.Money.From(value);
         
+
+        /// <summary>
+        /// User Interface for property 'CashNote'
+        /// </summary>
+        public TypeUserInterface? CashNoteUserInterface(NoxSolution solution)
```
rochar: now looking to the name.... @DanielHodgeGH  do y have a better suggestion?

CashNoteUiSettings
CashNoteUiOptions
CashNoteUi
:(

rochar (changes requested):

samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.VendingMachineMetadata.g.cs
rochar: > isRequired is on the same level as typeoptions and user interface. Please see below, it is taken from Cryptocash:
> 
> ```
>   - name: Name
>     description: Country's name
>     type: text
>     textTypeOptions:
>       minLength: 4
>       maxLength: 63
>     isRequired: true
>     userInterface:
>       showInSearchResults: optionalAndOnByDefault
>       canSort: true
>       canSearch: true
>       canFilter: true
>       showOnCreateForm: true
>       showOnUpdateForm: true
> ```
> 
> For the second option - yes, I totally agree that it's better to use userInterface params on the generation phase not from the UI app.

So IsRequired is not exposed in the metadata?

If that is true the Generator will not need metadata it can do it without it.... we can figure this better on the generator implementation phase

src/Nox.Generator/Domain/EntityMeta.template.cs
anna-naumova25: updated

src/Nox.Generator/Domain/EntityMeta.template.cs
anna-naumova25: updated

src/Nox.Generator/Domain/EntityMeta.template.cs
anna-naumova25: updated

samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Meta.BankNoteMetadata.g.cs
anna-naumova25: I stopped on CashNoteUiOptions as it correlates with the TypeOptions we already have in Metadata

rochar (approved):

PR #871, opened 2023-09-28
Issues/851 extend solution version and platfrom

### LINKED ISSUE #851: [Nox.Solution] Create PlatformId for Integration Event usage
related to #799 

# PlatformId And Version

For message identification a uri prefix will be used and configured in the Nox Solution:

1. Create a Property called PlatformId in the Solution root (next to the name)
    1. Required, but if not defined  default to solution name (so the user may not set it explicitly)
    2. Can not be set to null or empty 
    3. Its a string type
 1. Create a Property called Version (Version type) in the Solution root (next to the name)  
     2. Required, but if not defined  default 1.0 (so the user may not set it explicitly)  
     4.  Can not be set to null or empty

AndreyDegtyarev:
Introduced PlatformId and Version property in Solution class.

### DIFF
--- a/samples/.nox/design/cryptocash.solution.nox.yaml
+++ b/samples/.nox/design/cryptocash.solution.nox.yaml
@@ -6,10 +6,6 @@
 
 name: Cryptocash
 
-platformId: "https://www.nox.com"
-
-version: "1.0"
-
 description: A sample solution for the imaginary Crypocash business. 
 
 overview: >
--- a/samples/.nox/design/cryptocash.solution.nox.yaml
+++ b/samples/.nox/design/cryptocash.solution.nox.yaml
@@ -6,6 +6,10 @@
 
 name: Cryptocash
 
+platformId: "https://www.nox.com"
+
+version: "1.0"
+
 description: A sample solution for the imaginary Crypocash business. 
 
 overview: >
--- a/src/Nox.Solution/Models/Solution.cs
+++ b/src/Nox.Solution/Models/Solution.cs
@@ -19,15 +19,18 @@ public class Solution : DefinitionBase
     [Pattern(RegexConstants.SolutionNamePattern)]
     public string Name { get; set; } = null!;
 
-    [Title("The message identification Platform Id.")]
-    [Description("For message identification a PlatformId  will be used and configured in the Nox Solution. Required, but if not defined default to solution name (so the user may not set it explicitly).")]
-    public string PlatformId { get => _platformId ?? Name; internal set => _platformId = value; }
+    [Title("Platform Identifier. Used to build a unique Uri.")]
+    [Description("Identify a Platform, that is a set of different services. Use to produce a unique Uri, by encoding the provided value.")]
+    public string PlatformId
+    {
+        get => _platformId ?? Name;
+        internal set => _platformId = value;
+    }
 
-    [Required]
-    [Title("The version of the NOX solution.")]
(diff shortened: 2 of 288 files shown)


src/Nox.Solution/Models/Solution.cs
```diff
@@ -10,11 +11,23 @@ namespace Nox.Solution;
...
-    public string Name { get; internal set; } = null!;
+    [Pattern(RegexConstants.SolutionNamePattern)]
+    public string Name { get; set; } = null!;
+
+    [Title("The message identification Platform Id.")]
+    [Description("For message identification a PlatformId  will be used and configured in the Nox Solution. Required, but if not defined default to solution name (so the user may not set it explicitly).")]
+    public string PlatformId { get => _platformId ?? Name; internal set => _platformId = value; }
+
+    [Required]
+    [Title("The version of the NOX solution.")]
```
rochar:
```suggestion
    [Title("The version of the NOX solution. Expected a Semantic Version format.")]
```

src/Nox.Solution/Models/Solution.cs
```diff
@@ -10,11 +11,23 @@ namespace Nox.Solution;
...
+    [Description("Required, but if not defined default 1.0 (so the user may not set it explicitly).")]
```
rochar:
```suggestion
    [Description("Required, but if not defined defaults to 1.0 .")]
```

src/Nox.Solution/Models/Solution.cs
```diff
@@ -10,11 +11,23 @@ namespace Nox.Solution;
...
+    [Pattern(RegexConstants.SolutionVersionPattern)]
+    public string Version { get; internal set; } = null!;
```
rochar: The default value is not being set?

src/Nox.Solution/Models/Solution.cs
```diff
@@ -10,11 +11,23 @@ namespace Nox.Solution;
...
+
     [Required]
     [Title("The short name for the solution. Contains no spaces.")]
     [Description("The name of the NOX solution, application or service. This value is used extensively by the NOX tooling and libraries and should ideally be unique within an organisation.")]
-    [Pattern(@"^[^\s]*$")]
-    public string Name { get; internal set; } = null!;
+    [Pattern(RegexConstants.SolutionNamePattern)]
+    public string Name { get; set; } = null!;
+
+    [Title("The message identification Platform Id.")]
```
rochar:
```suggestion
    [Title("Platform Identifier. Used to build a unique Uri")]
```

src/Nox.Solution/Models/Solution.cs
```diff
@@ -10,11 +11,23 @@ namespace Nox.Solution;
...
+    [Description("For message identification a PlatformId  will be used and configured in the Nox Solution. Required, but if not defined default to solution name (so the user may not set it explicitly).")]
```
rochar:
```suggestion
    [Description("Identify a Platform, that is a set of different services. Use to produce a unique Uri, by encoding the provided value.")]
```

samples/.nox/design/cryptocash.solution.nox.yaml
rochar, line 1: no need to change all yamls, you need to have a default value and provide a test that asserts if th version is empty in the yaml, the value 1.0 is returned

rochar (changes requested):

rochar (approved):

PR #888, opened 2023-09-29
MP-9237 Example domain event handler created for Country entity

msadikkoseaupa:
Example domain event handler created for Country entity under Nox.ClientApi.Tests

Also Meta data code files updated with the latest version.


src/Nox.ClientApi.Tests/Application/DomainEventHandlers/CountryCreatedDomainEventHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+    
+    /// <summary>
+    /// Handles the domain event asynchronously.
+    /// </summary>
+    /// <param name="notification">The <see cref="CountryCreated"/> notification.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>A task representing the asynchronous operation.</returns>
+    public async Task Handle(CountryCreated notification, CancellationToken cancellationToken)
+    {
+        await _semaphoreSlim.WaitAsync(cancellationToken);
```
gjorgji-grgovski-authoritypartners: Why do we need a Semaphore here?
Also, we should have some real-case scenarios and custom logic inside the Domain Event handle.
Please merge this handler with the other ones: CountryCreatedEventHandler, CountryUpdatedEventHandler.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.ClientApi.Tests/Application/DomainEventHandlers/CountryCreatedDomainEventHandler.cs
msadikkoseaupa: Domain event handler test updated with real case scnenarios as desired. Folder structure will be updated with another pr https://github.com/NoxOrg/Nox.Generator/pull/893 which is from @dejobratic

src/Nox.ClientApi.Tests/Tests/Application/Messaging/DomainEventTests.cs
```diff
@@ -0,0 +1,47 @@
...
+                Population = 100_000_001,
+            };
+
+            // Act
+            var result = await PostAsync<CountryCreateDto, CountryDto>(CountriesControllerName, createDto);
+
+            //Assert
+            result.Should().NotBeNull();
+
+            (await MassTransitTestHarness.Published.Any<Nox.Messaging.NoxMessageRecord<CountryPopulationHigherThan100M>>()).Should().BeTrue();
```
gjorgji-grgovski-authoritypartners: You should be testing the CountryCreated event and not the Integration Event.

gjorgji-grgovski-authoritypartners: Also, please add a test for the other events like Updated, and Deleted.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Generator/Application/ServiceCollectionExtensions.template.cs
rochar, line 1: ok, like we saw we need to revert this file when you update this branch with main

src/Nox.ClientApi.Tests/Tests/Application/Messaging/DomainEventTests.cs
rochar: this can be a little bit challenging because we are doing System.Testing here, and mocking the IMediator, a core component of Nox,  and preserving the behavior can be difficult

Maybe we can have a different approach, and do more "business test",  for example build a scenario where when a country is created with a certain property value in the client application you subscribe to the countrycreated domain event and you update some property other of the country, the you can do a getbyid from the api for that country and assert that is the expected value .... so basically the DomainEvent was raised, the Notification handler was called and the bossiness rule was executed

@gjorgji-grgovski-authoritypartners what do you think?

rochar (changes requested):

gjorgji-grgovski-authoritypartners (approved):

src/Nox.ClientApi.Tests/Tests/Application/DomainEventHandlers/StoreOwnerCreatedDomainEventHandlerTests.cs
```diff
@@ -0,0 +1,56 @@
+using ClientApi.Application.Dto;
+using FluentAssertions;
+using Xunit.Abstractions;
+
+namespace ClientApi.Tests.Application.DomainEventHandlers;
+
+public class StoreOwnerCreatedDomainEventHandlerTests: NoxWebApiTestBase
+{
+    private const string StoreOwnersControllerName = "api/storeowners";
```
rochar, line 9: there is a new EndPoints class with constant that you can use

rochar (approved):

PR #889, opened 2023-09-29
887 Update Post when <RelationName>Id is set create the relation

### LINKED ISSUE #887: [Nox.Generator] Update Post when <RelationName>Id is set create the relation
For ZeroOrOne or ExactlyOne

1. Add to \<Entity\>CreateDto the \<RelationshipName\>Id (Not required and the user can set the navigation property)
2.  If is a single key we should have the primitive type has underline type 
3. if is a compose key we should have the keyDto type has underline type 
4. In the CreateCommand if the \<RelationName\>Id is set we need to get the Entity from the Db and associated it to the created entity, calling CreateRef<RelationName>

This is just for relation not OwnedRelationships

anna-naumova25:


src/Nox.ClientApi.Tests/Tests/Controllers/StoresControllerTests.cs
```diff
@@ -143,6 +144,142 @@ public async Task Get_StoreOwnerOdataQuery_ReturnOwner()
...
+            response.Should().NotBeNull();
+            response!.License.Should().NotBeNull();
+            response!.License!.Id.Should().Be(licensePostResponse!.Id);
+            response!.License!.Issuer.Should().Be(licenseName);
+        }
+        #endregion
+
+        #region POST Entity with Invalid RelationshipId /api/{EntityPluralName} => api/stores
+        [Fact]
+        public async Task Post_WithInvalidRelationshipId_CreatesEntityWithoutRelationship()
```
anna-naumova25: @rochar  if LicenseId has invalid Id then Store will be created anyway, License will be null
Is it what we want? Or do we want to throw an exception?

src/Nox.ClientApi.Tests/Tests/Controllers/StoresControllerTests.cs
```diff
@@ -143,6 +144,142 @@ public async Task Get_StoreOwnerOdataQuery_ReturnOwner()
...
+                .BeOfType<StoreDto>()
+                .Which.Id.Should().NotBeEmpty();
+            response.Should().NotBeNull();
+            response!.License.Should().BeNull();
+        }
+        #endregion
+
+        #region POST Entity with Deleted RelationshipId /api/{EntityPluralName} => api/stores
+        [Fact]
+        public async Task Post_WithDeletedRelationshipId_ShouldNotCreateRefToRelatedEntity()
```
anna-naumova25, line 229: @rochar 
I've just realized that we don't check in CreateRef endpoints. So if the user tries to create a ref between the existing and the deleted entity, the reference will be created. Should we have a ticket to fix it?

Also I'm not sure what happens if we get the entity, expand relationships and some of them are removed (do we filter them out?).

rochar: That function receives an entity, so that entity is new or already exists, so we should not need to verify if is delete, the caller should not set a deleted entity, this should not be possible since delete only exists in a database level for auditing

I have created this issue #890 so we do not need to care about this when querying entities or dtos. This will also take care of expanding the relations

samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands.CreateBookingCommand.g.cs
```diff
@@ -70,22 +70,50 @@ public virtual async Task<BookingKeyDto> Handle(CreateBookingCommand request, Ca
 		OnExecuting(request);
 
 		var entityToCreate = _entityFactory.CreateEntity(request.EntityDto);
-		if(request.EntityDto.BookingForCustomer is not null)
+		if(request.EntityDto.BookingForCustomerId is not null)
+		{
+			var relatedKey = CreateNoxTypeForKey<Customer, Nox.Types.AutoNumber>("Id", request.EntityDto.BookingForCustomerId);
+			var relatedEntity = await _dbContext.Customers.FindAsync(relatedKey);
+			if(relatedEntity is not null && relatedEntity.DeletedAtUtc == null)
```
rochar: we can remove the `relatedEntity.DeletedAtUtc == null` validation ticket #890 will solve this
thks

regarding your question if the entity does not exists we should throw an exception, do create a specific exception to throw , for example: RelatedEntityNotFoundException(request.EntityDto.BookingForCustomerId)

rochar (changes requested):
looks good, lets just refine the validation

samples/Cryptocash.Api/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands.CreateBookingCommand.g.cs
anna-naumova25: Updated for both points

src/Nox.ClientApi.Tests/Tests/Controllers/StoresControllerTests.cs
rochar: we want to throw a custom exception

rochar (approved):

PR #896, opened 2023-10-02
Compile tests were fixed for windows environment

msadikkoseaupa:


tests/Nox.Generator.Tests/Common/CompileTests.cs
```diff
@@ -94,6 +94,12 @@ private static IEnumerable<MetadataReference> GetReferences()
         {
```
rochar, line 94: all the three foreach and the dllFiles calculation could be encapsulated in a single method

rochar (approved):
thanks for fixing this

gjorgji-grgovski-authoritypartners (approved):

PR #903, opened 2023-10-03
MP-9352 Getting UserName and SystemName from Custom http Headers

msadikkoseaupa:


src/Nox.Lib/Application/Providers/DefaultSystemProvider.cs
```diff
@@ -1,9 +1,24 @@
...
+    public DefaultSystemProvider(IHttpContextAccessor httpContextAccessor) =>
+        _httpContextAccessor = httpContextAccessor;
+
+    /// <summary>
+    /// Get the system name from the X-System-Name header
+    /// </summary>
+    /// <returns></returns>
+    public string GetSystem()
+    {
+        var result = _httpContextAccessor.HttpContext?.Request.Headers["X-System-Name"].ToString();
```
rochar: I would suggest setting "X-System-Name" in a constant

src/Nox.Lib/Application/Providers/DefaultSystemProvider.cs
```diff
@@ -1,9 +1,24 @@
...
+        result ??= string.Empty;
```
rochar: Not sure what would be better, Empty? "NotFound"? Not Available? @gjorgji-grgovski-authoritypartners any suggestion?
```suggestion
        result ??= string.Empty;
```

src/Nox.Lib/Application/Providers/DefaultUserProvider.cs
rochar, line 1: same comments that were done on the SystemProvider

rochar (changes requested):

src/Nox.Lib/Application/Providers/DefaultSystemProvider.cs
gjorgji-grgovski-authoritypartners: The Header is supposed to be populated by the Kong Gateway, ideally, we would not have it as an empty, but in that case, I would still keep it as an empty string.
The thing here is that I think we should keep these DefaultProviders and create new ones like HeaderSystemProvider and HeaderUserProvider.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Lib/Application/Providers/DefaultSystemProvider.cs
rochar: this one needs to be the default setup (not name), because will be the one used for most of the services

We can Keep the DefaultProviders but not set them has defaults lol => or like @msadikkoseaupa  is proposing we fall back to Empty if not found, otherwise we should raise an exception

src/Nox.Lib/Application/Providers/DefaultSystemProvider.cs
msadikkoseaupa: Desired changes were applied

src/Nox.Lib/Application/Providers/DefaultSystemProvider.cs
msadikkoseaupa: If not found, returns "N/A" as we decided

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #904, opened 2023-10-03
Feature/890 noxlib implemente sofdelete filtering at the dbcontext

### LINKED ISSUE #890: [Nox.Lib] Implemente SofDelete Filtering at the DbContext
1. On Dto an Entity DbContext foll all IAuditable Entities add HasQueryFilter
```
      modelBuilder.Entity<Movie>()
            .HasQueryFilter(x => x.DeletedAtUtc == null);
```

1. Remove current filters in generated queries

anna-naumova25:


src/Nox.Generator/Infrastructure/Persistence/DbContext.template.cs
```diff
@@ -107,6 +108,15 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
         }
     }
 
+    private void ConfigureAuditable(ModelBuilder modelBuilder)
```
anna-naumova25, line 111: @rochar the solution we talked about yesterday doesn't work on the runtime. It is because the types in Lambda function is System.RuntimeType and not System.Type as it is expected by HasQueryFiler

`modelBuilder.ForEntitiesOfType<IEntityAuditDeleted>(builder => builder.HasQueryFilter((IEntityAuditDeleted p) => p.DeletedAtUtc == null));`

So I tried many different options but it seems to be the only one working.

rochar (approved):
looks good

PR #909, opened 2023-10-04
[Nox.Generator] Patch Endpoint - Delta<Entity> on Swagger

### LINKED ISSUE #902: [Nox.Generator] Patch Endpoint  - Delta<Entity> on Swagger
Our Patch endpoints are using the Delta to define the input parameter, example:

 ```
public virtual async Task<ActionResult<BookingDto>> Patch([FromRoute] System.Guid key, [FromBody] Delta<BookingDto> booking)
 {
     if (!ModelState.IsValid)
     {
         return BadRequest(ModelState);
     }
...
}
```

On swagger this is generating a incorrect schema:

Image

some reference: https://stackoverflow.com/a/65735078, investigate if there is a better way

### PR DESCRIPTION
**Pull request by AndreyDegtyarev**
Provided operation filter to substitute schema reference that contains Delta<>

### CODE REVIEW

File: samples/Cryptocash.Api/Infrastructure/DeltaSchemaFilter.cs
Comment (rochar) [inline samples/Cryptocash.Api/Infrastructure/DeltaSchemaFilter.cs:1]: @AndreyDegtyarev we need to implement this as a Nox Feature, making it transparent for our customers!

**Review by rochar — CHANGES_REQUESTED** [review]

File: samples/Cryptocash.Api/Program.cs [hunk 1]
```diff
@@ -1,33 +1,24 @@
...
 // Add services to the container.
 builder.Services.AddControllers();
 
-// Learn more about configuring Swagger/OpenAPI at https://aka.ms/aspnetcore/swashbuckle
 builder.Services.AddEndpointsApiExplorer();
-builder.Services.AddSwaggerGen();
-
-builder.AddNox();
 
+builder.AddNox(opts => opts.WithSwagger());
```
rochar, line 10: IMO this should be the default behavior of AddNox, we should allow to disable swagger instead

samples/Cryptocash.Api/Program.cs
```diff
@@ -1,33 +1,24 @@
...
-app.UseNox();
+var isDevelopment = app.Environment.IsDevelopment();
 
-// Configure the HTTP request pipeline.
-if (app.Environment.IsDevelopment())
-{
-    app.UseSwagger();
-    app.UseSwaggerUI();
-}
+app.UseNox(useSwagger: isDevelopment);
```
rochar, line 16: the same here.... this logic should be the default behavior of Nox

rochar (approved):
looks good, I'm merging it, and we can discuss my suggestion and do a new PR is necessary

PR #910, opened 2023-10-04
Add full namespace to all generated entities

### LINKED ISSUE #876: [Nox.Generator] add full namespace to all generated code
to avoid compilation issues common names, for example:

1. The Solution name is the same as an Entity
2. The Project Namespace is the same as an Entity

both cases should work properly

ILoveMyTortoise:
closes #876


rochar (dismissed):
looks good

anna-naumova25 (dismissed):

PR #916, opened 2023-10-04
feature 620 initialcommit for Add function simple nox types

### LINKED ISSUE #620: [Nox.Ui] Cryptocash.Ui Initial Add Entity Form functionality on Ui so that we can agree what to generate from config - initially Nox simple types

DanielHodgeGH:
PR for initial Cryptocash.Ui ADD function form using Nox simple types (MacAddress, PublicIp, SerialNumber)

ADD function also needs Compound Nox Types and List Selection but these handled on separate tickets - also needs to finish off POST to endpoint but this is on separate ticket as well


anna-naumova25 (approved):
It looks good to me in general. I'd like to talk about some Properties and Classes names, but we can discuss it when we start generating real file.

rochar (approved):

PR #920, opened 2023-10-06
MP-9440 - Generate Documentation for Integration Events

dejobratic:


gjorgji-grgovski-authoritypartners (dismissed):

gjorgji-grgovski-authoritypartners (dismissed):

src/Nox.Docs/Extensions/ScribanTemplateExtensions.cs
```diff
@@ -47,6 +50,14 @@ private static ScriptObject CreateScriptObject(Dictionary<string, object> model)
         var scriptObj = new ScriptObject();
         scriptObj.Import(model, renamer: member => member.Name, filter: null);
```
rochar: @dejobratic I think we could reuse the `NoxSolutionBridge` that is on Nox.Generator
I would be ok to add a dependency to Nox.Generator

Even this code   

```
 var scriptObj = new ScriptObject();
        scriptObj.Import(model, renamer: member => member.Name, filter: null);
```

it's also being used in the TemplateCodeBuilder.GenerateSourceCode we could probably add to NoxSolutionBridge (maybe rename this), and use in both sides

BTW we are creating a TemplateFileBuilder, in the feature we need to reassess if make sense the diverged path we are taking in both of the sides

rochar (changes requested):
looks good, one comment to review

rochar (approved):

PR #927, opened 2023-10-09
[Nox.Generators] Update ETag missing error message

### LINKED ISSUE #923: [Nox.Generators] Update ETag missing error message
Talk to @Mornevanzyl for more details

AndreyDegtyarev:


src/Nox.Lib/Extensions/HttpRequestExtensions.cs
```diff
@@ -1,21 +1,29 @@
...
-            rawEtag = encodedEtag.Tag.Trim('\"');
+            var rawEtag = encodedEtag.Tag.Trim('\"');
+            if (Guid.TryParse(rawEtag, out var etag))
+            {
+                return etag;
+            }
         }
 
-        return System.Guid.TryParse(rawEtag, out var etag) ? etag : null;
+        throw new ConcurrencyException("ETag has incorrect format.");
```
rochar:
```suggestion
        throw new ConcurrencyException("ETag is not well-formed.");
```

src/Nox.Lib/Middlewares/NoxExceptionHanderMiddleware.cs
```diff
@@ -31,7 +30,7 @@ public async Task InvokeAsync(HttpContext httpContext)
         {
             await HandleTypeValidationExceptionAsync(httpContext, ex);
         }
-        catch(ConcurrencyException ex)
+        catch (ConcurrencyException ex)
         {
             await HandleConcurrencyExceptionAsync(httpContext, ex);
```
rochar: this should be the status code if ETag is outdated 

//
// Summary:
//     Equivalent to HTTP status 412. System.Net.HttpStatusCode.PreconditionFailed indicates
//     that a condition set for this request failed, and the request cannot be carried
//     out. Conditions are set with conditional request headers like If-Match, If-None-Match,
//     or If-Unmodified-Since.
PreconditionFailed = 412,

src/Nox.Lib/Extensions/HttpRequestExtensions.cs
```diff
@@ -1,21 +1,29 @@
...
 {
-    public static System.Guid? GetDecodedEtagHeader(this HttpRequest request)
+    public static Guid GetDecodedEtagHeader(this HttpRequest request)
     {
         var ifMatchValue = request.Headers.IfMatch.FirstOrDefault();
-        string? rawEtag = ifMatchValue;
+
+        if (string.IsNullOrEmpty(ifMatchValue))
+        {
+            throw new ConcurrencyException("ETag is empty. ETag should be provided via the If-Match HTTP Header.");
```
rochar: we need to distinguish this exceptions, because if the etag is not well formed, the status code should be 400

if the Etag condition fails should be 412, needs to be clear to the user what is the type of exception

Should we add a status code to our exceptions ,if its not 500 that, we want to raise ? and set on an interface to be handled by the middleware INoxHandledException

src/Nox.Lib/Middlewares/NoxExceptionHanderMiddleware.cs
rochar: we can improve the middleware, avoid all this if, and adding some metadata to expected exceptions, see my comment above?

rochar (changes requested):

rochar (approved):

ILoveMyTortoise (approved):

PR #930, opened 2023-10-09
773 Create initial PR for generating razor files in Nox

### LINKED ISSUE #773: [Nox.Ui] Create initial PR for generating razor files in Nox
When trying to generate a file with razor extension from Nox was that current Roslyn API that is used for generation is only able to generate .cs files. Please see some issues and solutions I’ve found.
https://github.com/dotnet/roslyn/issues/57608
https://github.com/dotnet/roslyn/blob/v3.8.0/src/Compilers/Core/Portable/SourceGeneration/AdditionalSourcesCollection.cs#L63

Try one of the following:
- Generate files using existing flow and then rename created .razor.cs files to .razor
- use the fix from the link above and write files with File.WriteAllText(…)

anna-naumova25:


src/Nox.Generator/Application/ServiceCollectionExtensionsGenerator.cs
```diff
@@ -6,7 +6,11 @@ namespace Nox.Generator.Application;
 
 internal class ServiceCollectionExtensionsGenerator : INoxCodeGenerator
 {
-    public NoxGeneratorKind GeneratorKind => NoxGeneratorKind.None;
+    public NoxGeneratorKind GeneratorKind =>
```
anna-naumova25: Should be refactored in the future for proper generation of ServiceCollectionExtensionsGenerator.cs

src/Nox.Generator/Ui/Components/NavigationMenu.template.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using Microsoft.AspNetCore.Components;
+
+namespace {{codeGeneratorState.UiNameSpace}}.Components;
+
+public class NavigationMenu : ComponentBase
+{
+    {{- for entity in entities }}
+    
+    public string {{entity.PluralName}}Icon { get; set; } = "";
```
anna-naumova25, line 13: Will update it with icons path later

src/Nox.Generator/Ui/Components/NavigationMenuRazor.template.cs
```diff
@@ -0,0 +1,13 @@
+﻿/*
```
anna-naumova25, line 1: Will uncomment it in the next PR with razor file generation

DanielHodgeGH (dismissed):

src/Nox.Generator/Ui/Components/NavigationMenuGenerator.cs
rochar, line 1: so can you also add to the PR the generated crypto cash ui file?

src/Nox.Generator/Application/ServiceCollectionExtensionsGenerator.cs
rochar: IMO you should set has Application, so the Application layer is responsible to services registration, and so on.... actually this one is already in the proper namespace

rochar (changes requested):

src/Nox.Generator/Application/ServiceCollectionExtensionsGenerator.cs
anna-naumova25: have updated the PR to set the Application NoxGeneratorKind only

rochar (approved):

Issue #932, opened 2023-10-09
[Nox.Ui] Add Nox to Cryptocash.UI


anna-naumova25:
1. Move all prepared for generation files to a new folder (ie Generated.Prepared)
2. Add Nox to the Cryprocash.UI
3. Make sure .razor files are generated and can be used it other views.

FYI @rochar @DanielHodgeGH

PR #933, opened 2023-10-09
931 Update Nox.Generator to allow saving files with System.IO.File

### LINKED ISSUE #931: [Nox.Generator] [Nox.Ui] Update Nox.Generator to allow saving files with System.IO.File
Based on #773 investigation, extend Nox.Generator to allow saving files with System.IO.File instead of using current context flow.

FYI @rochar @DanielHodgeGH

anna-naumova25:


src/Nox.Generator/Common/TemplateCodeBuilder.cs
```diff
@@ -136,10 +150,23 @@ private void GenerateSourceCode(string template, object model, string sourceFile
...
-        _context.AddSource(sourceFileName,
-            SourceText.From(sourceText,
-            Encoding.UTF8));
+
+        if (string.IsNullOrEmpty(_fileExtension) || _fileExtension == "cs")
+        {
+            _context.AddSource($"{sourceFileName}.g.cs", SourceText.From(sourceText, Encoding.UTF8));
+        }
+        else
+        {
```
anna-naumova25: This is a very rough implementation. We should probably configure output path on the client side, maybe in appsetings.json

src/Nox.Generator/Common/TemplateCodeBuilder.cs
rochar, line 1: @anna-naumova25 having a new generator/template we should have the new generated files, in this case in Crypto Cash UI, to be easy to validate the teplate

src/Nox.Generator/Common/TemplateCodeBuilder.cs
rochar: @anna-naumova25  lets think about composition or inheritance to solve this, so we keep 2 generators the simple and with only the specific code.

src/Nox.Generator/Common/TemplateCodeBuilder.cs
```diff
@@ -136,10 +150,23 @@ private void GenerateSourceCode(string template, object model, string sourceFile
...
+#pragma warning disable RS1035 // Do not use APIs banned for analyzers
+            string outputFolder = "Generated.NonCs";
```
rochar:
```suggestion
            string outputFolder = "Nox.Generated";
```

rochar (changes requested):

src/Nox.Generator/Common/TemplateCodeBuilder.cs
anna-naumova25: Sure, I've created a separate ticket for it - #932 as there will be lots of changes. So I wanted to keep this PR simple and clear.

src/Nox.Generator/Common/TemplateCodeBuilder.cs
anna-naumova25: I've create TemplateCodeBuilderBase and two derived classes - TemplateCodeBuilder and TemplateCodeBuilderFile to separate the logic

src/Nox.Generator/Common/TemplateCodeBuilderFile.cs
rochar, line 1: looka much better now

rochar (approved):

ILoveMyTortoise (approved):

PR #946, opened 2023-10-10
MP-9447 Create function on Entity Add/Delete Owned Entity

msadikkoseaupa:


tests/Nox.Integration.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Domain.TestEntityOwnedRelationshipOneOrMany.g.cs
```diff
@@ -59,38 +59,66 @@ internal abstract partial class TestEntityOwnedRelationshipOneOrManyBase : Audit
...
+            throw new Exception($"The relationship cannot be deleted.");
+        SecondTestEntityOwnedRelationshipOneOrMany.Remove(relatedSecondTestEntityOwnedRelationshipOneOrMany);
+    }
+    
+    /// <summary>
+    /// Deletes all owned SecondTestEntityOwnedRelationshipOneOrMany entities.
+    /// </summary>
+    public virtual void DeleteAllRefToSecondTestEntityOwnedRelationshipOneOrMany()
+    {
+        if(SecondTestEntityOwnedRelationshipOneOrMany.Count() < 2)
```
msadikkoseaupa, line 118: Could you please pay extra attention to this part. I got it from Relationship implementation. However if relationship is one or many, delete all ref  should directly throw exception instead of checking count. What do you think? 
If you agree with me I can update both owned entity and relationship delete all ref implementations. 
@ggrgovski @dejobratic @rochar

src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -251,17 +251,92 @@ public virtual void ClearDomainEvents()
...
     /// </summary>
     {{- if relationship.Relationship == "ZeroOrMany" || relationship.Relationship == "OneOrMany"}}
-    public virtual List<{{relationship.Entity}}> {{relationship.Name}} { get; set; } = new();
-    {{- else}}
-     public virtual {{relationship.Entity}}{{if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Name}} { get; set; } = null!;
-    {{-end}}
-{{- end }}
+    public virtual List<{{relationship.Entity}}> {{relationship.Name}} { get; private set; } = new();
+	{{- else}}
+    public virtual {{relationship.Entity}}{{if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationship.Name}} { get; private set; }{{if relationship.Relationship == "ExactlyOne"}} = null!;{{end}}
```
gjorgji-grgovski-authoritypartners, line 265: Why do we need a check if it is "ExactlyOne" to set the default value to null?

src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -251,17 +251,92 @@ public virtual void ClearDomainEvents()
...
+    {{- if relationship.ShouldGenerateForeignOnThisSide && (relationship.Related.Entity.Keys | array.size) > 0 }}
```
rochar: this is used in normal relations, where we need to define where the relation is going to be generated.

owned entities, the relation is on the main entity always, and are fully manage in the main entity, Id is not needed. Do we need this change in the scope of this ticket?

rochar (changes requested):
looks good, one comment to review

gjorgji-grgovski-authoritypartners (dismissed):
Approved with a small comment

src/Nox.Generator/Domain/Entity.template.cs
gjorgji-grgovski-authoritypartners: Thanks @rochar, let's change it with this ticket.

tests/Nox.Integration.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Domain.TestEntityOwnedRelationshipOneOrMany.g.cs
rochar: I'm ok with this for now... until we have some real cases

 I have seen that even EF on the database level its does not enforce it, because, I Guess in many real cases although is mandatory in the initial state is not possible to have it.... for example (just guessing) I may create a workplace (physical one) in my domain, but until is ready for usage I cannot associate a Tenant although is mandatory to have one!

rochar: @msadikkoseaupa  I would suggest to create a specific exception (avoid using System.Exception) and implement the INoxHttpException, so you can set the status code that will be returned in this case...

rochar (changes requested):

rochar (changes requested):
thanks @msadikkoseaupa  good job

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #963, opened 2023-10-11
MP-9075 NOX - Refactor EntityControllerGenerator to use Scriban template - Migrating endpoints for Entity

emir-sehovic1:


src/Nox.Generator/Common/TemplateScriptsBridges/NoxSolutionBridge.cs
```diff
@@ -43,6 +45,17 @@ public static void AddFunctions(TemplateContext context, NoxSolution noxSolution
         scriptObject8.Import("IsValueType", new Func<string,
             bool>(type => Type.GetType(type)?.IsValueType ?? false));
 
+        var scriptObject9 = new ScriptObject();
```
rochar, line 48: see https://github.com/scriban/scriban/blob/master/doc/builtins.md#string-functions, thera are scriban native functions, we may remove this

src/Nox.Generator/Common/TemplateScriptsBridges/NoxSolutionBridge.cs
```diff
@@ -43,6 +45,17 @@ public static void AddFunctions(TemplateContext context, NoxSolution noxSolution
...
+        scriptObject9.Import("ToLowerFirstChar", new Func<string, string>(
+            input => input.ToLowerFirstChar()));
+
+        var scriptObject10 = new ScriptObject();
+        scriptObject10.Import("GetPrimaryKeys", PrimaryKeysQuery);
+
+        var scriptObject11 = new ScriptObject();
```
rochar: this is only related with API Generator and the `PrimaryKeysFromRoute `can be done directly on the template so its clear what we are generating

and the same for this function `PrimaryKeysQuery`

rochar: @emir-sehovic1  To keep it simple and move forward with the PR, lets add the result of this 2 function to the generator class, so its only available for this template

```
  new TemplateCodeBuilder(context, codeGeneratorState)
      .WithClassName($"{entity.PluralName}Controller")
      .WithFileNamePrefix("Presentation.Api.OData")
      .WithFileNameSuffix("Entity")
      .WithObject("entity", entity)
     .WithObject("PrimaryKeysQuery", GetPrimaryKeysQuery())
    ...
```

src/Nox.Generator/Common/TemplateScriptsBridges/NoxSolutionBridge.cs
rochar: that way we can remove the link to string build extensions

rochar (changes requested):
looks good @emir-sehovic1 , juts two comments to considerer

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #989, opened 2023-10-13
Feat/#950 is localized text type options validations

### LINKED ISSUE #950: [Nox.Solution] IsLocalized TextTypeOptions Validations
- relates to #907

The IsLocalized Property for TextTypeOptions must have the following validations 

1. IsLocalized  must be false if OwnedEntity
2. IsLocalized  must be false if Entity has a compound key

We are doing this because we are not implementing this features for now.

ILoveMyTortoise:
closes #950


ILoveMyTortoise:
Sorry for space changes

rochar (approved):

PR #990, opened 2023-10-13
Implement basic classes and test for new 'Enumeration' value object.

andresharpe:
- Added 'Enumeration' nox type
- Implemented value object
- Added new tests for the type
- Renamed 'Nox.Types.Common.Enumeration' -> 'Nox.Types.Common.SmartEnumeration' to avoid confusion
- Updated Nox.Solution with Enumeration and its EnumerationTypeOptions

### DIFF
--- a/src/Nox.Types.Abstractions/TypeOptions/EnumerationTypeOptions.cs
+++ b/src/Nox.Types.Abstractions/TypeOptions/EnumerationTypeOptions.cs
@@ -11,7 +11,7 @@ public class EnumerationTypeOptions : INoxTypeOptions
 {
     [Title("The optional default values for this Enum.")]
     [Description("The default evalues of this Enum. These values will automatically be created and available in the domain.")]
-    public IReadOnlyList<EnumerationValues>? Values { get; set; } 
+    public IReadOnlyList<EnumerationValues> Values { get; set; } = System.Array.Empty<EnumerationValues>();
 
     [Title("Whether the description of the Enum will be localizable.")]
     [Description("Whether the description of the Enum will be localizable for the supported solution cultures")]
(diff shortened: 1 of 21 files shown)


src/Nox.Types.Abstractions/TypeOptions/EnumerationTypeOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Nox.Types;
+
+[Title("Options for the enum value type.")]
+[Description("Options to control the behaviour of the Enum nox type.")]
+[AdditionalProperties(false)]
+public class EnumerationTypeOptions : INoxTypeOptions
+{
+    [Title("The optional default values for this Enum.")]
+    [Description("The default evalues of this Enum. These values will automatically be created and available in the domain.")]
+    public IReadOnlyList<EnumerationValues>? Values { get; set; }
```
rochar: I would prefer to have an Empty Collection as the default value, and avoid having a nullable property, and therefore avoid unnecessary null checks
```suggestion
    public IReadOnlyList<EnumerationValues>? Values { get; set; }  = Array.Empty<EnumerationValues>();
```

src/Nox.Types/Types/Enumeration/Enumeration.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// Creates a new instance of <see cref="Enumeration"/> using the specified <see cref="EnumerationTypeOptions"/>
+    /// </summary>
+    /// <param name="value">The Enum integer value to create the <see cref="Text"/> with</param>
+    /// <param name="options">The <see cref="EnumerationTypeOptions"/> containing constraints for the value object</param>
+    /// <returns>A new <see cref="Enumeration" instance./></returns>
+    /// <exception cref="ValidationException">If the enum value is invalid.</exception>
+    public static Enumeration From(string stringValue, EnumerationTypeOptions options)
+    {
+
+        var value = options.Values?.FirstOrDefault(o => o.Description.Equals(stringValue, StringComparison.InvariantCultureIgnoreCase))?.Id;
```
rochar: we should treat collections as non-nullable

rochar (changes requested):
looks good, I would just avoid nullable collections

rochar (approved):

PR #998, opened 2023-10-16
997 Create NoxFileGenerator

### LINKED ISSUE #997: [Nox.Generator] Split Code and Files generators
We need to split Code and Files generation into 2 different flows in order to create build Task that can be used in the Blazor Server Applications.

FYI @rochar

anna-naumova25:


src/Nox.Generator/NoxFileGenerator.cs
```diff
@@ -0,0 +1,171 @@
...
+using YamlDotNet.Serialization.NamingConventions;
+using YamlDotNet.Serialization;
+using Nox.Solution;
+using System.Reflection;
+
+namespace Nox.Generator;
+
+public class NoxFileGenerator
+{
+    private readonly List<string> _errors = new();
```
anna-naumova25, line 19: @rochar do we need a debug file with listed errors (like Generator.cs) for NoxFileGenerator?

src/Nox.Generator/NoxFileGenerator.cs
```diff
@@ -0,0 +1,171 @@
...
+
+    public void GenerateFiles()
+    {
+        try
+        {
+            if (TryGetGeneratorConfig(out var config) && TryGetNoxSolution(out var solution))
+            {
+                var codeGeneratorState = new NoxSolutionCodeGeneratorState(solution, Assembly.GetEntryAssembly()!);
+
+                var generatorFlows = new (NoxGeneratorKind GeneratorKind, bool IsEnabled)[]
```
anna-naumova25, line 41: This can be moved to a helper class if needed

anna-naumova25:
@rochar 

I've added FilesGeneratorTask in a new Nox.Generator.Tasks project. The reason for this is to use netstandard2.0 target framework for Tasks, please see the link - https://learn.microsoft.com/en-us/answers/questions/459914/custom-buld-task-when-used-in-csproject-throws-cao

The target will look like this if we use FilesGeneratorTask 
```
	<ItemGroup>
		<InputFiles Include="../.nox/design/**/*.yaml" />
		<InputFiles Include="generator.nox.yaml" />
		<OutputFiles Include="Nox.Generated" />
	</ItemGroup>

	<UsingTask TaskName="Nox.Generator.Tasks.FilesGeneratorTask" AssemblyFile="bin\Debug\net7.0\Nox.Generator.Tasks.dll" />

	<Target Name="GenerateFiles" BeforeTargets="Build" Inputs="@(InputFiles)" Outputs="@(OutputFiles)">
		<FilesGeneratorTask NoxYamlFiles="@(InputFiles)"></FilesGeneratorTask>
	</Target>
```

src/Nox.Solution/Nox.Solution.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>netstandard2.0;net7.0</TargetFrameworks>
```
anna-naumova25, line 4: Need to find the way how to use netstandard2.0 target only in Nox.Generator and keep net7.0 here

src/Nox.Generator/NoxFileGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+                    (NoxGeneratorKind.Infrastructure,config.Infrastructure),
+                    (NoxGeneratorKind.Presentation,config.Presentation),
+                    (NoxGeneratorKind.Application,config.Application),
+                    (NoxGeneratorKind.Ui,config.Ui)
+                }
+                .Where(x => x.IsEnabled)
+                .Select(x => x.GeneratorKind)
+                .ToArray();
+
+                //var generatorInstances = Assembly
```
anna-naumova25, line 54: Assembly .GetExecutingAssembly().GetTypes() is not compatible with netstandard2.0 for any reasons. It needs to be changed.

PR #1003, opened 2023-10-17
Localization Data Model - generated entity localized

### LINKED ISSUE #936: [Nox.Generator] - Localization Data Model - generated entity localized
- relates to #907

# Data Model

We will create an additional class model (EF) for each Entity that contains all localized fields, for example, for Country Entity, if the Description and ShortDescription are localized fields, then we will have:

```
  public class CountryLocalized: IEntityConcurrent
  {    
      public number CountryId { get; set; } // Primary & Foreign key
      public CultureCode CultureCode { get; set; } // Primary Key
      public text Description { get; set; }
      public text ShortDescription { get; set; }
  }
```
## Tasks

1. Create a new template : \<Entity\>Localized
2. Generate a new Model, for All Entities that have at least one localized field (exclude owned entities)
3. In the Entity namespace
4. Add the Entity key  - {KeyNameId} (For now support only entities with single key)
5. Add the CultureCode Property (will be set as Primary Key with the Entity Foreign Key)
6. Add an attribute for each localized field, with the proper Nox.Type
7. ADd IEntityConcurrent interface to support concurrency when updating

ILoveMyTortoise:
closes #936

### DIFF
--- a/src/Nox.Generator/Domain/EntitiesLocalizedGenerator.cs
+++ b/src/Nox.Generator/Domain/EntitiesLocalizedGenerator.cs
@@ -1,0 +1,49 @@
+﻿using Microsoft.CodeAnalysis;
+using Nox.Generator.Common;
+using Nox.Solution;
+using Nox.Types;
+using System.Linq;
+
+namespace Nox.Generator.Domain;
+
+internal class EntitiesLocalizedGenerator : INoxCodeGenerator
+{
+    public NoxGeneratorKind GeneratorKind => NoxGeneratorKind.Domain;
+
+    public void Generate(SourceProductionContext context, NoxSolutionCodeGeneratorState codeGeneratorState, GeneratorConfig config, string? projectRootPath)
+    {
+        context.CancellationToken.ThrowIfCancellationRequested();
+
+        if (codeGeneratorState.Solution.Domain is null) return;
+
+        foreach (var entity in codeGeneratorState.Solution.Domain.Entities)
+        {
+            // Currently skip owned and composite key entities
+            if (entity.IsOwnedEntity ||
+                entity.Keys.Count > 1)
+            {
+                continue;
+            }
+
+            var entityHasAnyAttributesToLocalize =
+                entity.Attributes?.Count > 0 &&
+                entity.Attributes.Any(x =>
+                    x.Type == NoxType.Text &&
+                    x.TextTypeOptions != null &&
+                    x.TextTypeOptions.IsLocalized);
+
+            if (!entityHasAnyAttributesToLocalize)
+            {
+                continue;
+            }
+
+            context.CancellationToken.ThrowIfCancellationRequested();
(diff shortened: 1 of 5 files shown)


src/Nox.Generator/Domain/EntitiesLocalizedGenerator.cs
```diff
@@ -0,0 +1,49 @@
...
+        foreach (var entity in codeGeneratorState.Solution.Domain.Entities)
+        {
+            // Currently skip owned and composite key entities
+            if (entity.IsOwnedEntity ||
+                entity.Keys.Count > 1)
+            {
+                continue;
+            }
+
+            var entityHasAnyAttributesToLocalize =
```
rochar: @ILoveMyTortoise  all null checks can be removed, collections must be empty
All the cases you have here should be fixed:

`var entityAttributesToLocalize = entity.AttributesWhere(x => x.Type == NoxType.Text && x.TextTypeOptions.IsLocalized).Select(attributes=>new {Entity = entity, Attributes= attributes});`

src/Nox.Generator/Domain/EntityLocalized.template.cs
```diff
@@ -0,0 +1,53 @@
...
+    public virtual {{key.EntityIdTypeOptions.Entity}} {{key.EntityIdTypeOptions.Entity}} { get; set; } = null!;
+
+    {{- else -}}
+
+    public Nox.Types.{{key.Type}} {{key.Name}} { get; set; } = null!;
+    {{- end}}
+{{- end }}
+
+      public Nox.Types.CultureCode? CultureCode { get; set; }
+{{- for attribute in entity.Attributes }}
```
rochar: (Just a suggestion) another option is to set the calculated entityHasAnyAttributesToLocalize to the template and iterate it, no need to test any condition
```suggestion
{{- for attribute in entityAttributesToLocalize  }}
```

rochar (changes requested):
looks good, juts a suggestion to consider

src/Nox.Generator/Domain/EntityLocalized.template.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public virtual {{key.EntityIdTypeOptions.Entity}} {{key.EntityIdTypeOptions.Entity}} { get; set; } = null!;
+
+    {{- else -}}
+
+    public Nox.Types.{{key.Type}} {{key.Name}} { get; set; } = null!;
+    {{- end}}
+{{- end }}
+
+      public Nox.Types.CultureCode? CultureCode { get; set; }
```
rochar: This will be always mandatory
```suggestion
      public Nox.Types.CultureCode CultureCode { get; set; }
```

rochar (changes requested):

rochar (approved):

PR #1037, opened 2023-10-25
MP-9653 - Add Accept-Language Header Provider

dejobratic:


rochar (approved):

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    private readonly string _defaultLanguage;
+    private readonly HashSet<string> _supportedLanguages;
+
+    public DefaultLanguageProvider(
+        NoxSolution solution,
+        IHttpContextAccessor httpContextAccessor)
+        : base(httpContextAccessor)
+    {
+        _defaultLanguage = solution.Application?.Localization?.DefaultCulture ?? "en-US"; // should we set the language or throw an exception?
```
Tsarar: Maybe just en?

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+        _supportedLanguages = solution.Application?.Localization?.SupportedCultures?.ToHashSet() ?? new HashSet<string>();
+    }
+
+    public override string HeaderName => "Accept-Language";
```
ILoveMyTortoise: maybe use HeaderNames.AcceptLanguage

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public override string DefaultHeaderValue => _defaultLanguage;
+
+    public string GetLanguage()
+    {
+        var language = GetHeaderValue();
+        language = language.Split(',').Select(x => x.Trim()).First();
```
ILoveMyTortoise: what if first is "fr;q=0.9". You need to also trim by ';'

tests/Nox.Lib.Tests/Application/Providers/DefaultLanguageProviderTests.cs
```diff
@@ -0,0 +1,101 @@
...
+    [InlineData("en")]
+    [InlineData("fr-CH")]
+    [InlineData("fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5")]
+    public void GetLanguage_WhenHeaderIsSetWithUnsupportedLanguage_ReturnsDefaultValue(string value)
+    {
+        // Arrange
+        var expected = "en-US";
+
+        var httpContext = new DefaultHttpContext();
+        httpContext.Request.Headers.Add("Accept-Language", value);
```
ILoveMyTortoise: same, can use const

tests/Nox.Lib.Tests/Application/Providers/DefaultLanguageProviderTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        var actual = _provider.GetLanguage();
+
+        // Assert
+        actual.Should().Be(expected);
+    }
+
+    [Theory]
+    [InlineData("fr-FR")]
+    [InlineData("fr-FR, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5")]
+    public void GetLanguage_WhenHeaderIsSet_ReturnsHeaderValue(string value)
```
ILoveMyTortoise: Add also a test for a scenario where the first language is unsupported but the second/third is supported.

ILoveMyTortoise (dismissed):

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
rochar: I think we just need a default for Application and Localization on the NoxSolution and avoid all null checks

for example: `public Application Application { get; internal set; } = new Application();`

rochar (changes requested):

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
rochar: for now lets keep this simple, a following ticket will take care of this

src/Nox.Abstractions/Application/ILanguageProvider.cs
gjorgji-grgovski-authoritypartners, line 1: This should be on the Presentations layer.

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
gjorgji-grgovski-authoritypartners, line 1: This should be on the Presentations layer.

src/Nox.Lib/Application/Providers/HttpHeaderValueProvider.cs
gjorgji-grgovski-authoritypartners, line 1: This should be on the Presentations layer.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
dejobratic: Should we also move the DefaultUserProvider and DefaultSystemProvider, since both of them also have HttpContextAccessor as dependency?

src/Nox.Abstractions/Application/ILanguageProvider.cs
dejobratic: Not sure if the interface should be on the presentation layer, since we could use the same interface with different implementations for other mediums (service bus etc.)?

tests/Nox.Lib.Tests/Application/Providers/DefaultLanguageProviderTests.cs
dejobratic: Applied.

tests/Nox.Lib.Tests/Application/Providers/DefaultLanguageProviderTests.cs
dejobratic: Applied.

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
dejobratic: Good catch. Applied.

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
dejobratic: Thanks for the suggestion. Applied.

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
dejobratic: I've implemented Ricardo's suggestion with default Application and Localization objects. Thanks for the answer anyways.

src/Nox.Lib/Application/Providers/HttpHeaderValueProvider.cs
dejobratic: Done.

src/Nox.Abstractions/Application/ILanguageProvider.cs
rochar, line 1: the definition should also be in Presentation/API

rochar (dismissed):

src/Nox.Abstractions/Application/ILanguageProvider.cs
rochar: But that needs to be a different contract, otherwise would be tricky to distinguish them?

probably you can call it something like `IHttpLaunguageProvider `to be very clear the intention and the scope?

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
ILoveMyTortoise: @dejobratic you can also add a test for it if you think it's relevant. Please resolve after you add/decide not to add it

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
rochar: yes you can do it, not mandatory for the PR

src/Nox.Abstractions/Application/ILanguageProvider.cs
dejobratic: Done.

dejobratic: Done.

src/Nox.Lib/Application/Providers/DefaultLanguageProvider.cs
dejobratic: Done.

rochar (approved):

ILoveMyTortoise (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #1050, opened 2023-10-25
[Nox.Solution] Improve Ensure Keys and Attributes #627

### LINKED ISSUE #627: [Nox.Solution] Improve Ensure Keys and Attributes
Nox.Solution.Entity ensures keys and attributes EnsureAttributesByName and EnsureKeyByName , this can be simplified if called them in ApplyDefaults, and we can remove null checks

need to fix tests that are not calling ApplyDefaults

AndreyDegtyarev:


src/Nox.Solution/Models/Domain/Entities/Entity.cs
rochar, line 1: @AndreyDegtyarev  now in Ensure methods you can remove the lock and replace the concurrent by a normal dictionary
Apply defaults will be called only once

rochar (changes requested):

rochar (approved):

PR #1053, opened 2023-10-26
Added ui config to relationships

andresharpe:
Allow UI hints to be added to relationship declarations.

### DIFF
--- a/src/Nox.Solution/Models/Domain/Entities/Relationships/EntityRelationship.cs
+++ b/src/Nox.Solution/Models/Domain/Entities/Relationships/EntityRelationship.cs
@@ -1,5 +1,6 @@
 ﻿using Humanizer;
 using Nox.Solution.Extensions;
+using Nox.Types;
 using Nox.Types.Schema;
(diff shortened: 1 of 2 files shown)


DanielHodgeGH (approved):

src/Nox.Solution/Models/Domain/Entities/Relationships/EntityRelationship.cs
```diff
@@ -32,6 +33,8 @@ public class EntityRelationship : DefinitionBase
     [Description("The name of the target entity that this entity relates to.")]
     public string Entity { get; internal set; } = null!;
 
+    public TypeUserInterface? UserInterface { get; internal set; }
```
rochar, line 36: Just a suggestion to make TypeUserInterface? UserInterface non nullable => public TypeUserInterface UserInterface { get; internal set; } = new TypeUserInterface()
 
And therefore avoid null checks, and spreading out default values everywhere?

rochar (changes requested):

PR #1054, opened 2023-10-26
1030 noxgenerator enhance put verb for entity relationships

### LINKED ISSUE #1030: [Nox.Generator] Enhance Put verb for Entity Relationships
When Updating an Entity The related entities must be set using their Id's, for single (zero or one) and multiple (zero or Many). If null means that the relations will be deleted.

Support only Related Entities with single Keys.

anna-naumova25:


src/Nox.Generator/Domain/Entity.template.cs
```diff
@@ -205,6 +205,19 @@ public virtual void ClearDomainEvents()
         {{- end }}
     }
 
+    {{- if relationship.WithMultiEntity }}
+
+    public virtual void UpdateAllRefTo{{relationship.Name}}(List<{{relationship.Entity}}> related{{relationship.Entity}})
```
rochar: just a suggestion
```suggestion
    public virtual void UpdateRefTo{{relationship.Name}}(List<{{relationship.Entity}}> related{{relationship.Entity}})
```

rochar (dismissed):
thks good job

PR #1070, opened 2023-10-31
NOX-3 Update CreateDto and Commands

anna-naumova25:


src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
```diff
@@ -71,8 +71,8 @@ public abstract class {{className}}Base : IEntityDto<DomainNamespace.{{entity.Na
     [System.Text.Json.Serialization.JsonIgnore]
     public virtual {{relationship.Entity}}CreateDto? {{relationship.Name}} { get; set; } = default!;
     {{- else }}
-    {{#Simplify Avoid complexity by not allowing circular dependency between dtos
-    #TODO Allow to set a list of Ids for a *ToMany relations}}
+    public virtual List<{{relationship.ForeignKeyPrimitiveType}}> {{relationship.Name}}Id { get; set; } = new();
+    {{#Simplify Avoid complexity by not allowing circular dependency between dtos}}
     [System.Text.Json.Serialization.JsonIgnore]
```
rochar, line 76: @anna-naumova25  we need to remove the JsonIgnore so its shown on Swagger (we can do it on next PR)

rochar: we also need to set it on Odata configuration, for example

`builder.EntityType<CountryDto>().ContainsMany(e => e.PhysicalWorkplaces)`

src/Nox.ClientApi.Tests/Tests/Controllers/CountriesControllerTests.cs
rochar, line 1: see also this test `Post_WithManyRelatedEntities_Success`, I think we can remove the skip?

rochar (changes requested):
looks very good, just small refinements

rochar (approved):

PR #1076, opened 2023-11-02
NOX-36 - PartialUpdate<Entity>Command save to Localized table

dejobratic:
https://regusit.atlassian.net/browse/NOX-36


src/Nox.Generator/Application/Factories/EntityLocalizedFactory.template.cs
```diff
@@ -18,36 +18,52 @@ internal partial class {{className}} : {{className}}Base
 
 internal abstract class {{className}}Base : IEntityLocalizedFactory<{{localizedEntityName}}, {{entity.Name}}Entity, {{entity.Name}}UpdateDto>
 {
-    public virtual {{localizedEntityName}} CreateLocalizedEntity({{entity.Name}}Entity entity, CultureCode cultureCode)
+    public virtual {{localizedEntityName}} CreateLocalizedEntity({{entity.Name}}Entity entity, CultureCode cultureCode, bool withAttributes = true)
```
rochar: For better readability:

```suggestion
    public virtual {{localizedEntityName}} CreateLocalizedEntity({{entity.Name}}Entity entity, CultureCode cultureCode, bool copyEntityAttributes= true)
```

src/Nox.Solution/Extensions/EntityExtensions.cs
```diff
@@ -51,7 +51,11 @@ public static bool TryGetRelationshipByName(this Entity entity, NoxSolution solu
         }
     }
 
-    public static IEnumerable<NoxSimpleTypeDefinition> GetAttributesToLocalize(this Entity entity) => entity
-        .Attributes
-        .Where(x => x.IsLocalized);
+    public static IEnumerable<NoxSimpleTypeDefinition> GetAttributesToLocalize(this Entity entity)
+        => entity.Attributes.Where(x => x.IsLocalized).Select(property =>
```
rochar, line 55: IMO this is a kind of an hack, 
We should change the code according to be a localized or not, for example in EntitLocalized.template should set everything to optional and not use the IsRequired, because that is the rule

rochar: and in the NoxDatabaseConfigurator. ConfigureLocalizedAttributes you can clone and change the is required, because there it will be explicit that we are configuring the db and set all fields to optional

src/Nox.Generator/Application/Dto/EntityLocalizedDto.template.cs
```diff
@@ -34,9 +34,9 @@ internal partial class {{className}}
     public System.String {{codeGeneratorState.LocalizationCultureField}} { get; set; } = default!;
 {{ for attribute in entityAttributesToLocalize }}
     /// <summary>
-    /// {{attribute.Description}} ({{if attribute.IsRequired}}Required{{else}}Optional{{end}}).
+    /// {{attribute.Description}} (Optional).
     /// </summary>
-    public {{attributeType attribute}}{{ if !attribute.IsRequired}}?{{end}} {{attribute.Name}} { get; set; }{{if attribute.IsRequired}} = default!;{{end}}
+    public {{attributeType attribute}}? {{attribute.Name}} { get; set; }
```
rochar, line 39: this is good we are being explicitly here!

src/Nox.Generator/Application/Commands/PartialUpdateOwnedCommand.template.cs
```diff
@@ -70,7 +70,7 @@ internal abstract class PartialUpdate{{entity.Name}}For{{parent.Name}}CommandHan
 			return null;
 		}
 
-		EntityFactory.PartialUpdateEntity(entity, request.UpdatedProperties);
+		EntityFactory.PartialUpdateEntity(entity, request.UpdatedProperties, DefaultCultureCode);
```
rochar, line 73: Owned Entities need also to support localized text fields
We are creating a ticket for handling this

rochar (changes requested):
looks good, please catch up with @ggrgovski  you just need to do the change regarding the property name "withAttributes"

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #1083, opened 2023-11-03
NOX-48 Update Endpoint for $ref

anna-naumova25:


src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -25,6 +25,9 @@ public class Entity : DefinitionBase
     [YamlIgnore]
     private ConcurrentDictionary<string, NoxSimpleTypeDefinition>? _keysByName;
 
+    [YamlIgnore]
+    private ConcurrentDictionary<EntityRelationship, string>? _relationshipPublicNamesByRelationship;
```
rochar: @anna-naumova25  no need to be a ConcurrentDictionary since apply defaults is called only once on the validation phase, there will no be multiple threads calling this method

rochar (changes requested):
looks good just one comment to review

rochar (approved):

PR #1097, opened 2023-11-10
NOX-8 - Allow to override Language Header via Query String

dejobratic:
https://regusit.atlassian.net/browse/NOX-8

### DIFF
--- a/samples/Cryptocash.Ui/Nox.Generated/Generator.g.cs
+++ b/samples/Cryptocash.Ui/Nox.Generated/Generator.g.cs
@@ -2,7 +2,7 @@
 
 #nullable enable
 
-// Using Nox version 7.0.84.0
+// Using Nox version 7.0.86.0
 
 /* Found files ->
 - cryptocash.application.nox.yaml
--- a/schemas/entityRelationship.json
+++ b/schemas/entityRelationship.json
@@ -70,8 +70,8 @@
       ]
     },
     "canNavigate": {    
-      "title": "Determines whether this side of the relationship is exposed in the generated code and ODATA endpoints.",
-      "description": "This boolean controls whether this side of the relationship is exposed in the generated code and ODATA endpoints.",
+      "title": "Define the Entity on this relationship side contains a Navigation Property to the related Entity.",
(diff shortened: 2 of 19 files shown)


src/Nox.Lib/Presentation/Api/Providers/HttpHeaderValueProvider.cs
rochar, line 1: to be honest, IMO I do not see any added value on this interfaces IHttpHeaderValueProvider and IHttQueryParam... is just an abstraction in top of existing IHttpContextAccessor that creates an extra level of complexity

rochar (approved):
Looks good, although we should think about the abstractions we are creating in top of the IHttpContextAccessor

src/Nox.Lib/Presentation/Api/Providers/HttpHeaderValueProvider.cs
dejobratic: I've added them because I wanted the same logic to be present in all places where query/header extraction is needed - if it's null or empty we default to null. And also makes the unit testing easier. But, I can remove them if required.

rochar: I do not want you to remove it :)  Just want to have a debate it

What you are saying do not justify to create a new service, that could be done with an extension method?

gjorgji-grgovski-authoritypartners: @rochar @dejobratic I agree with both of you :)
I would suggest we merge this PR and we can discuss this quickly 3 of us, will agree on the best practices and we can modify if needed further.
However it was a good discussion, and potentially we can define some good practices here.

gjorgji-grgovski-authoritypartners (approved):

src/Nox.Lib/Presentation/Api/Providers/HttpHeaderValueProvider.cs
dejobratic: IMO, isolated services are easier to mock and test, but we don't have a complex logic here so extension methods should work. Whatever we agree on :)

PR #1102, opened 2023-11-10
NOX-169 AutoNumber Fields can not be set on the Create and Update Dto

emir-sehovic1:


src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
```diff
@@ -48,6 +48,9 @@ public abstract class {{className}}Base : IEntityDto<DomainNamespace.{{entity.Na
 {{- end }}
 
 {{- for attribute in entity.Attributes }}
+    {{- if attribute.Type == "AutoNumber" -}}
```
gjorgji-grgovski-authoritypartners: Extend SimpleType with the "Create" flag and consult it here.

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):
good job

PR #1105, opened 2023-11-10
NOX-199 New EndPoint to Update Related Entity from Entity Api

anna-naumova25:
@rochar I want to confirm 2 things before this one is merged:
1. We agreed to return Updated() but it cannot go without the entity as a parameter. Right now it is returned empty Ok(). Let me know if we need to change it to Updated(relatedEntity) please. It will require to make an additional DB call to get the updated entity.
2. Let's discuss the behavior of how the relationships of the related entity will be updated in this endpoint. Cause I found some cases when they can be overwritten.


src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData/WorkplacesController.g.cs
```diff
@@ -104,6 +104,29 @@ public async Task<ActionResult> DeleteRefToCountry([FromRoute] System.UInt32 key
...
     }
     
+    public virtual async Task<ActionResult<CountryDto>> PutToCountry(System.UInt32 key, [FromBody] CountryUpdateDto country)
+    {
+        if (!ModelState.IsValid)
+        {
+            return BadRequest(ModelState);
+        }
+        
+        var related = (await _mediator.Send(new GetWorkplaceByIdQuery(_cultureCode, key))).Select(x => x.Country).SingleOrDefault();
```
rochar, line 114: we could improve the efficiency of the query  by doing something like 
```suggestion
        var related = (await _mediator.Send(new GetWorkplaceByIdQuery(_cultureCode, key))).Select(x => x.Country.Id).SingleOrDefault();
```

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData/WorkplacesController.g.cs
```diff
@@ -184,6 +207,30 @@ public async Task<ActionResult> DeleteRefToTenants([FromRoute] System.UInt32 key
...
     
+    [HttpPut("api/Workplaces/{key}/Tenants/{relatedKey}")]
+    public virtual async Task<ActionResult<TenantDto>> PutToTenantsNonConventional(System.UInt32 key, System.Guid relatedKey, [FromBody] TenantUpdateDto tenant)
+    {
+        if (!ModelState.IsValid)
+        {
+            return BadRequest(ModelState);
+        }
+        
+        var related = (await _mediator.Send(new GetWorkplaceByIdQuery(_cultureCode, key))).Select(x => x.Tenants).SingleOrDefault()?.SingleOrDefault(x => x.Id == relatedKey);
```
rochar: Would this work, or similiar to?
```suggestion
        var related = (await _mediator.Send(new GetWorkplaceByIdQuery(_cultureCode, key))).Select(x => x.Tenants).SingleOrDefault()?.Any(x => x.Id == relatedKey);
```

rochar (changes requested):
looks good two small improvements

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData/WorkplacesController.g.cs
anna-naumova25: I'd leave it as is for this case to cover composite key scenarios. Otherwise it gets to complicated like this:

```
var related = (await _mediator.Send(new GetWorkplaceByIdQuery(_cultureCode, key)))
.Select(x => new { x.Country?.Id1, x.Country?.Id2} )
.SingleOrDefault();

if(related == null || related.Id1 == null || related.Id2 == null)
return NotFound();
```

And Country is not always nullable so we have to check when generating `Country?.Id` which brings extra complexity.

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData/WorkplacesController.g.cs
anna-naumova25: This one is updated

rochar (approved):

PR #1107, opened 2023-11-11
define Nox.Solution configuration for setting the Api route prefix

rochar:


rochar:
@andresharpe 
We need to distinguish between validators and the concept of applying defaults.

Otherwise, it becomes difficult to manage the order in which a validation or a default is applied. Additionally, it is complex to apply defaults using other Solution properties because defaults may not have been set yet.

So first Validate the Solution, second step apply defaults!

src/Nox.Solution/Models/Infrastructure/Infrastructure.cs
```diff
@@ -7,23 +8,27 @@ namespace Nox.Solution;
...
 public class Infrastructure : DefinitionBase
-{
-    [Required]
+{    
     // These descriptors should be moved to the class when the generator is fixed
     [Title("The definition namespace for persistance settings pertaining to a Nox solution.")]
     [Description("Defines settings pertinent to solution persistence here. These include database, event source, search and cache servers.")]
     [AdditionalProperties(false)]
-    public Persistence Persistence { get; internal set; } = new();
+    public Persistence? Persistence { get; internal set; }
```
rochar, line 16: Persistence was required only if Infrastructure existed, now we can have Parts of Infrastructure without Persistence

src/Nox.Solution/NoxSolutionBuilder.cs
```diff
@@ -66,11 +66,19 @@ public NoxSolution Build()
         {
             ResolveAndValidateFilesAndContent();
 
-            if (_yamlFilesAndContent == null) return new NoxSolution();
+            if (_yamlFilesAndContent == null)
+            {
+                var noxSolution = new NoxSolution();
+                // Should we not validate the Default Solution?
```
rochar, line 72: IMO we should explicitly validate the default NoxSolution if it's supposed to be used!

rochar:
I will merge until we understand the impact of Nox.Yaml

PR #1135, opened 2023-11-17
Feature/integration solution changes

jan-schutte:


rochar:
@jan-schutte  your PR has 1375 files changes, I guess you have commited some generated files that were deleted on build, can y see this first?

Also can you Move Nox.Integration.Tests to folder ./tests?

rochar (changes requested):
see my comment please
thks

jan-schutte:
Hi @rochar I've fixed the files as requested

src/Nox.Solution/Enums/Application/Integration/IntegrationSourceAdapterType.cs
```diff
@@ -0,0 +1,10 @@
+namespace Nox;
+
+public enum IntegrationSourceAdapterType
```
rochar, line 3: I would suggest just to add the adapter that we are supporting at the moment
and comment the other ones
to avoid creating wrong expectations
and to be clear what we are supporting at the moment

for now is DatabaseQuery only?

src/Nox.Solution/Enums/Application/Integration/IntegrationTargetAdapterType.cs
```diff
@@ -0,0 +1,11 @@
+namespace Nox;
+
+public enum IntegrationTargetAdapterType
```
rochar, line 3: same comment here!

src/Nox.Solution/Validation/Application/Integration/Target/IntegrationTargetTableOptionsValidator.cs
rochar, line 1: this will probably conflict with what Andre is doing, maybe we can not commit this now and integrate validation when Nox.Yaml is fully integrated

rochar (changes requested):
can you move Nox.Integration.Tests to ./tests

rochar (dismissed):

PR #1147, opened 2023-11-20
NOX-280 Create <OwnedEntity>UpsertDto, replace CreateDto and UpdateDto

anna-naumova25:


src/Nox.Generator/Application/Dto/EntityUpsertDtoGenerator.cs
```diff
@@ -0,0 +1,58 @@
...
+using Nox.Generator.Common;
+using Nox.Solution;
+using Nox.Types;
+using Nox.Types.Extensions;
+
+namespace Nox.Generator.Application.Dto;
+
+internal class EntityUpsertDtoGenerator : INoxCodeGenerator
+{
+    public NoxGeneratorKind GeneratorKind => NoxGeneratorKind.Domain;
```
anna-naumova25: @rochar It's used NoxGeneratorKind.Domain for all the Dto models. Is it correct? Shouldn't we use NoxGeneratorKind.Application?

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -27,18 +29,18 @@
 
 namespace {{codeGeneratorState.ApplicationNameSpace}}.Factories;
 
-internal abstract class {{className}}Base : IEntityFactory<{{entity.Name}}Entity, {{entity.Name}}CreateDto, {{entity.Name}}UpdateDto>
+internal abstract class {{className}}Base : IEntityFactory<{{entity.Name}}Entity, {{entityCreateDto}}, {{entityUpdateDto}}>
```
anna-naumova25, line 32: @rochar As we have only one factory per entity, it means we'll have to use UpsertDto instead of CreateDto and UpdateDto for Owned entities in all the commands. That is what we discussed, just wanted to confirm that at the end we can remove CreateDto and UpdateDto for owned entities.

src/Nox.Generator/Application/Dto/EntityUpsertDtoGenerator.cs
gjorgji-grgovski-authoritypartners: Yes to me it looks like the Dtos and Factories have the wrong NoxGeneratorKind, they should all have the Application and not Domain.

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
rochar: I was looking to the IFactory definition, and I think it fits, since we are saying for specific entities that the update and create dto is the same

src/Nox.Generator/Application/Dto/EntityUpsertDtoGenerator.cs
```diff
@@ -0,0 +1,58 @@
...
+      string? projectRootPath
+      )
+    {
+        NoxSolution solution = codeGeneratorState.Solution;
+        context.CancellationToken.ThrowIfCancellationRequested();
+
+        if (solution.Domain is null || !solution.Domain.Entities.Any())
+            return;
+
+        foreach (var entity in codeGeneratorState.Solution.Domain!.Entities.Where(entity => entity.IsOwnedEntity))
```
gjorgji-grgovski-authoritypartners:
```suggestion
        foreach (var entity in solution.Domain!.Entities.Where(entity => entity.IsOwnedEntity))
```

src/Nox.Generator/Application/Dto/EntityUpsertDtoGenerator.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            var attributes = entity.Attributes ?? Enumerable.Empty<NoxSimpleTypeDefinition>();
```
gjorgji-grgovski-authoritypartners:
```suggestion
            var attributes = entity.Attributes;
```

You don't need to have a null check, because it would instantiate this property with a default value which is an empty array

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
```diff
@@ -45,11 +45,13 @@ public partial class {{className}}Base: EntityDtoBase, IEntityDto<DomainNamespac
 {{- # for relationship in entity.Relationships # see NOX-237 to enable relationships in UpdateDto}}
 {{- for relationship in entity.OwnedRelationships}}
 	{{- relationshipName = GetNavigationPropertyName entity relationship }}
-    {{- if relationship.Relationship == "ZeroOrOne" || relationship.Relationship == "ExactlyOne"}}
     /// <summary>
     /// {{entity.Name}} {{relationship.Description}} {{relationship.Relationship}} {{relationship.EntityPlural}}
     /// </summary>
-    public virtual {{relationship.Entity}}UpdateDto{{- if relationship.Relationship == "ZeroOrOne"}}?{{end}} {{relationshipName}} { get; set; } = null!;
+    {{- if relationship.Relationship == "ZeroOrOne" || relationship.Relationship == "ExactlyOne"}}
```
gjorgji-grgovski-authoritypartners:
```suggestion
    {{- if relationship.WithSingleEntity}}
```

src/Nox.Generator/Application/Dto/EntityUpsertDtoGenerator.cs
rochar: NoxGeneratorKind.Application

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
```diff
@@ -93,9 +93,9 @@ public abstract class {{className}}Base : IEntityDto<DomainNamespace.{{entity.Na
     /// {{entity.Name}} {{relationship.Description}} {{relationship.Relationship}} {{relationship.EntityPlural}}
     /// </summary>
     {{- if relationship.Relationship == "ZeroOrMany" || relationship.Relationship == "OneOrMany"}}
```
gjorgji-grgovski-authoritypartners:
```suggestion
    {{- if relationship.WithMultiEntity}}
```

or maybe align with the UpdateDto and use the relationship.WithSingleEntity condition and in the else generate the list property.

rochar (dismissed):

gjorgji-grgovski-authoritypartners:
Can we also exclude OwnedEntities in the EntityUpdateDtoGenerator and EntityCreateDtoGenerator, so that we do not generate anymore the UpdateDto/CreateDto for owned entities and that we can also identify it they are still used somewhere else?

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Generator/Application/Dto/EntityUpsertDtoGenerator.cs
anna-naumova25: Updated

src/Nox.Generator/Application/Dto/EntityUpsertDtoGenerator.cs
anna-naumova25: Updated

src/Nox.Generator/Application/Dto/EntityUpdateDto.template.cs
anna-naumova25: Updated

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
anna-naumova25: Updated

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):

PR #1156, opened 2023-11-22
NOX-238 Delete all related endpoint must execute a single command

AndreyDegtyarev:
Delete all related endpoint must execute a single command

### DIFF
--- a/src/Nox.Generator/Application/Commands/DeleteRelatedCommand.template.cs
+++ b/src/Nox.Generator/Application/Commands/DeleteRelatedCommand.template.cs
@@ -15,12 +15,12 @@ using {{entity.Name}}Entity = {{codeGeneratorState.DomainNameSpace}}.{{entity.Na
 
 namespace {{codeGeneratorState.ApplicationNameSpace}}.Commands;
 
-public partial record DeleteAll{{relationship.Name}}For{{parent.Name}}Command({{parent.Name}}KeyDto ParentKeyDto) : IRequest <bool>;
+public partial record DeleteAll{{GetNavigationPropertyName parent relationship}}For{{parent.Name}}Command({{parent.Name}}KeyDto ParentKeyDto) : IRequest <bool>;
 
 
-internal partial class DeleteAll{{relationship.Name}}For{{parent.Name}}CommandHandler : DeleteAll{{relationship.Name}}For{{parent.Name}}CommandHandlerBase
+internal partial class DeleteAll{{GetNavigationPropertyName parent relationship}}For{{parent.Name}}CommandHandler : DeleteAll{{GetNavigationPropertyName parent relationship}}For{{parent.Name}}CommandHandlerBase
 {
-	public DeleteAll{{relationship.Name}}For{{parent.Name}}CommandHandler(
+	public DeleteAll{{GetNavigationPropertyName parent relationship}}For{{parent.Name}}CommandHandler(
         AppDbContext dbContext,
 		NoxSolution noxSolution)
 		: base(dbContext, noxSolution)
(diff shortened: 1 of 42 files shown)


src/Nox.Generator/Application/Commands/DeleteRelatedCommand.template.cs
```diff
@@ -0,0 +1,90 @@
...
+using Nox.Types;
+using Nox.Application.Factories;
+using {{codeGeneratorState.PersistenceNameSpace}};
+using {{codeGeneratorState.DomainNameSpace}};
+using {{codeGeneratorState.ApplicationNameSpace}}.Dto;
+using {{entity.Name}}Entity = {{codeGeneratorState.DomainNameSpace}}.{{entity.Name}};
+
+namespace {{codeGeneratorState.ApplicationNameSpace}}.Commands;
+
+public partial record DeleteAll{{relationship.Name}}For{{parent.Name}}Command({{parent.Name}}KeyDto ParentKeyDto) : IRequest <bool>;
```
anna-naumova25: It should be used {{GetNavigationPropertyName parent relationship}} instead of {{relationship.Name}}

src/Nox.Generator/Application/Commands/DeleteRelatedCommand.template.cs
```diff
@@ -0,0 +1,90 @@
...
+			{{- relationshipName = GetNavigationPropertyName parent relationship }}
+			var related = parentEntity.{{relationshipName}};
+			if (related == null)
+			{
+				return false;
+			}
+			
+			foreach(var relatedEntity in related)
+			{
+				DbContext.{{relationship.EntityPlural}}.Remove(relatedEntity);
```
anna-naumova25, line 73: There should be different logic for Auditable entities:

```
{{- if (entity.Persistence?.IsAudited ?? true) }}
DbContext.Entry(entity).State = EntityState.Deleted;
{{-else-}}
DbContext.{{entity.PluralName}}.Remove(entity);
{{- end}}
```

src/Nox.Generator/Application/Commands/DeleteRelatedCommandGenerator.cs
```diff
@@ -0,0 +1,34 @@
...
+internal class DeleteRelatedCommandGenerator : ApplicationEntityDependentGeneratorBase
+{
+    protected override void DoGenerate(SourceProductionContext context, NoxCodeGenConventions codeGeneratorState, IEnumerable<Entity> entities)
+    {
+        var templateName = @"Application.Commands.DeleteRelatedCommand";
+        foreach (var entity in entities.Where(x => !x.IsOwnedEntity))
+        {
+            context.CancellationToken.ThrowIfCancellationRequested();
+
+            foreach (var ownedRelation in entity.Relationships.Where(x => x.WithMultiEntity))
```
anna-naumova25: Why is it called `ownedRelation` if it is a normal relation?

anna-naumova25:
@AndreyDegtyarev @rochar do we really want to have these additional commands for each relationship? 

@AndreyDegtyarev can't we just reuse the existing DeleteById command and send the list of Ids as the request? 

Also I don't see where this new command is used in the code. 

And we'll probably need some tests in ClientApiTests to cover it.

anna-naumova25 (changes requested):

rochar:
> @AndreyDegtyarev @rochar do we really want to have these additional commands for each relationship?
> 
> @AndreyDegtyarev can't we just reuse the existing DeleteById command and send the list of Ids as the request?
> 
> Also I don't see where this new command is used in the code.
> 
> And we'll probably need some tests in ClientApiTests to cover it.

@AndreyDegtyarev  the idea was just to change the current delete command instead of receiving a single entity, would receive a list => We are deleting a entity via a related end point the command should not care about it

this is not being called anywhere!!!

rochar:
we will pick this later

PR #1158, opened 2023-11-22
NOX-281 Add Validation for the commands where <OwnedEntity>UpsertDto is used

anna-naumova25:


src/Nox.Generator/Presentation/Api/OData/EntityControllerGenerator.cs
```diff
@@ -399,6 +369,15 @@ private static void GenerateChildrenPatch(NoxSolution solution, EntityRelationsh
             }}           
         }}");
         code.AppendLine();
+
+        if (child.Keys.Any())
+        {
+            code.AppendLine($"if(!updateProperties.ContainsKey(\"{child.Keys[0].Name}\"))");
+            code.StartBlock();
+            code.AppendLine($"return BadRequest(\"{child.Keys[0].Name} is required.\");");
```
anna-naumova25: Will need to apply latest changes for Exception here once they are merged

src/Nox.Generator/Presentation/Api/OData/EntityControllerGenerator.cs
```diff
@@ -304,20 +304,9 @@ private static void GenerateChildrenPut(NoxSolution solution, EntityRelationship
...
-        }
-        else
-        {
-            code.AppendLine($"[HttpPut(\"{solution.Infrastructure.Endpoints.ApiRoutePrefix}/{parent.PluralName}/{PrimaryKeysAttribute(parent)}/{relationshipName}/{PrimaryKeysAttribute(child, "relatedKey")}\")]");
-            code.AppendLine($"public virtual async Task<ActionResult<{child.Name}Dto>> PutTo{child.PluralName}NonConventional(" +
-                $"{GetPrimaryKeysRoute(parent, solution, attributePrefix: "")}, " +
-                $"{GetPrimaryKeysRoute(child, solution, "relatedKey", "")}, " +
-                $"[FromBody] {child.Name}UpsertDto {child.Name.ToLowerFirstChar()})");
-        }
+        code.AppendLine($"public virtual async Task<ActionResult<{child.Name}Dto>> PutTo{relationshipName}(" +
```
rochar, line 307: we should use the related entity name? to make it consistent with "normal" Relations
for owned we do not have the case of relating to the same entity

And we should keep the plural if ToMany and singular if to one?

We have other issue to rename the OwnedEntitites Navigation properties, and think we should align also the endpoints

rochar: Another question, do you think its easy to migrate this to the template for owned, now that you are changing this?

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -71,7 +74,7 @@ public virtual void PartialUpdateEntity({{entity.Name}}Entity entity, Dictionary
             {{- if !IsNoxTypeCreatable key.Type || key.Type == "Guid" -}}
                 {{ continue; -}}
             {{- end }}
-        entity.{{key.Name}} = {{entity.Name}}Metadata.Create{{key.Name}}(createDto.{{key.Name}}{{if entity.IsOwnedEntity}}!{{end}});
+        entity.{{key.Name}} = {{entity.Name}}Metadata.Create{{key.Name}}(createDto.{{key.Name}}{{if entity.IsOwnedEntity}}.NonNullValue<{{keyType key}}>(){{end}});
```
rochar, line 77: now looking to this, we can not have an owned used as a Foreign key, if this is possible in Nox.Solution that is incorrect.

@anna-naumova25  lets discuss this use case

rochar (changes requested):
looks good, lets just discuss some points

src/Nox.Generator/Presentation/Api/OData/EntityControllerGenerator.cs
```diff
@@ -304,20 +304,9 @@ private static void GenerateChildrenPut(NoxSolution solution, EntityRelationship
         var isSingleRelationship = relationship.WithSingleEntity();
         var relationshipName = parent.GetNavigationPropertyName(relationship);
```
rochar: rename to navigationPropertyName

rochar (changes requested):

src/Nox.Generator/Presentation/Api/OData/EntityControllerGenerator.cs
anna-naumova25: done

src/Nox.Generator/Presentation/Api/OData/EntityControllerGenerator.cs
anna-naumova25: done

rochar (approved):

PR #1164, opened 2023-11-23
Feature/integration handler generator

jan-schutte:


rochar (dismissed):
looks good, I guess you did not commited the generated file?

PR #1178, opened 2023-11-28
Integration - target watermark

jan-schutte:


src/Nox.Solution/Models/Application/Integrations/Target/IntegrationTargetWatermark.cs
```diff
@@ -0,0 +1,18 @@
...
+[Description("This section lists the column(s) on the target data used to determine whether a record is new or has been updated.")]
+[AdditionalProperties(false)]
+public class IntegrationTargetWatermark
+{
+    [Title("Sequential key column used for watermark purpose.")]
+    [Description("Specify one or more sequential key columns to be compared to source data to determine if a record exits in the target.")]
+    public IReadOnlyList<string>? SequentialKeyColumns { get; internal set; }
+    
+    [Title("Column(s) in datetime format used for watermark purpose.")]
+    [Description("List od datetime columns to be compared to source data to determine if data has changed.")]
```
rochar:
```suggestion
    [Description("List of datetime columns to be compared to source data to determine if data has changed.")]
```

src/Nox.Integration/Services/NoxIntegration.cs
```diff
@@ -30,8 +30,10 @@ public NoxIntegration(ILogger logger, Solution.Integration definition)
         Description = definition.Description;
         MergeType = definition.MergeType;
         TransformType = definition.TransformationType;
-        if (definition.Source.Watermark == null) return;
-        var watermark = definition.Source.Watermark;
+
```
rochar: Consider encapsulating this logic in a methods, since is not just assigning variables

rochar (dismissed):

PR #1181, opened 2023-11-29
Allow different signatures for Http verbs on dynamic Api middleware

andresharpe:
- Maintain separate api patterns per verb in middleware
- Allows unique match per verb for same endpoint


rochar (approved):
thanks

PR #1184, opened 2023-11-29
NOX-282 Use <OwnedEntity>UpsertDto in PUT to Main Entity to manage owned entities

anna-naumova25:


src/Nox.Lib/Extensions/FluentApiExtensions.cs
```diff
@@ -15,4 +15,12 @@ public static IEnumerable<T> ForEach<T>(this IEnumerable<T> array, Action<T> act
             act(i);
         return array;
     }
+    public static bool HasAtLeastOneItem<T>(this List<T> list)
+    {
+        return list != null && list.Count > 0;
```
rochar: Just needs to find the first item
```suggestion
        return list != null && list.Count.Any();
```

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands/UpdateCountryCommand.g.cs
```diff
@@ -56,6 +57,18 @@ internal abstract class UpdateCountryCommandHandlerBase : CommandBase<UpdateCoun
...
+			.Where(x => x.Id != null)
+			.Select(x => ClientApi.Domain.CountryLocalNameMetadata.CreateId(x.Id.NonNullValue<System.Int64>()));
+		foreach(var ownedEntity in entity.CountryLocalNames)
+		{
+			if(!keysToUpdateCountryLocalNames.Any(x => x == ownedEntity.Id))
+				DbContext.Entry(ownedEntity).State = EntityState.Deleted;
+		}
+		await DbContext.Entry(entity).Reference(x => x.CountryBarCode).LoadAsync();
+		if(entity.CountryBarCode is not null)
+			DbContext.Entry(entity.CountryBarCode).State = EntityState.Deleted;
```
rochar: May be I miss something, we are deleting the non existing, but where do we add the new ones (Owned)?

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands/UpdateStoreCommand.g.cs
```diff
@@ -56,6 +57,9 @@ internal abstract class UpdateStoreCommandHandlerBase : CommandBase<UpdateStoreC
 		{
 			return null;
 		}
+		await DbContext.Entry(entity).Reference(x => x.EmailAddress).LoadAsync();
+		if(entity.EmailAddress is not null)
+			DbContext.Entry(entity.EmailAddress).State = EntityState.Deleted;
```
rochar: the same here, not sure where we are adding the new one/update the ones to be updated

rochar: I see now that we are doing it on the factory.... I think we need to consolidate this responsibility, somewhere, maybe in the factory, lets discuss this :)

rochar (changes requested):

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands/UpdateCountryCommand.g.cs
rochar: we should also consider start using DbContext.Remove() its more meaningful then  setting a state

src/Nox.Lib/Extensions/FluentApiExtensions.cs
anna-naumova25: Done

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands/UpdateCountryCommand.g.cs
anna-naumova25: Done

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands/UpdateStoreCommand.g.cs
anna-naumova25: Implemented via IRepository as discussed

src/Nox.Lib/Extensions/FluentApiExtensions.cs
```diff
@@ -15,4 +15,12 @@ public static IEnumerable<T> ForEach<T>(this IEnumerable<T> array, Action<T> act
             act(i);
         return array;
     }
+    public static bool HasAtLeastOneItem<T>(this List<T> list)
```
gjorgji-grgovski-authoritypartners, line 18: This is a ListExtenstion and not FluentApiExtensions, maybe we can move it to another class, to be more clear?

src/Nox.Lib/Infrastructure/Persistence/EntityDbContextBase.cs
```diff
@@ -8,14 +8,13 @@
...
-using System.Reflection.Emit;
 using Nox.Types.EntityFramework.Types;
 using Nox.Types.EntityFramework.Abstractions;
-using static Microsoft.EntityFrameworkCore.DbLoggerCategory.Database;
+using Nox.Extensions;
 
 namespace Nox.Infrastructure.Persistence
 {
-    public abstract class EntityDbContextBase : DbContext
+    public abstract class EntityDbContextBase : DbContext, IRepository
```
gjorgji-grgovski-authoritypartners: Usually, the DbContext and the repository are separate concerns with different responsibilities.
I would suggest that we create a Repository class that implements IRepository, which would inject DbContext in the constructor.

src/Nox.Generator/Application/Commands/UpdateCommand.template.cs
```diff
@@ -1,4 +1,7 @@
-﻿﻿﻿// Generated
+﻿{{- func keyType(key)
```
gjorgji-grgovski-authoritypartners: I don't see usage of this function

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
```diff
@@ -190,17 +197,75 @@ private void PartialUpdateEntityInternal({{entity.Name}}Entity entity, Dictionar
...
+			{{- if ownedRelationship.WithSingleEntity }}
+            if(entity.{{navigationName}} is not null)
+                {{ownedRelationship.Entity}}Factory.UpdateEntity(entity.{{navigationName}}, updateDto.{{navigationName}}, cultureCode);
+            else
+			    entity.CreateRefTo{{navigationName}}({{ownedRelationship.Entity}}Factory.CreateEntity(updateDto.{{navigationName}}));
+			{{- else # WithMultiEntity }}
+			var updated{{navigationName}} = new List<{{codeGeneratorState.DomainNameSpace}}.{{ownedRelationship.Entity}}>();
+			foreach(var ownedUpsertDto in updateDto.{{navigationName}})
+			{
+				if(ownedUpsertDto.{{key.Name}} is null)
```
gjorgji-grgovski-authoritypartners, line 232: I don't think this would work for Guid, because you can set the Guid if you like.

src/Nox.Generator/Application/Commands/UpdateCommand.template.cs
```diff
@@ -107,4 +120,25 @@ private async Task UpdateLocalizedEntityAsync({{entity.Name}}Entity entity, {{en
...
+public class Update{{entity.Name}}Validator : AbstractValidator<Update{{entity.Name}}Command>
+{
+    public Update{{entity.Name}}Validator()
+    {
+		{{- for ownedRelationship in entity.OwnedRelationships }}
+			{{- if ownedRelationship.WithMultiEntity }}
+				{{- relationshipName = GetNavigationPropertyName entity ownedRelationship }}
+				{{- key = ownedRelationship.Related.Entity.Keys | array.first }}
+					{{- if IsNoxTypeCreatable key.Type }}
+		RuleFor(x => x.EntityDto.{{relationshipName}})
```
gjorgji-grgovski-authoritypartners, line 134: This can be RuleForEach, which would then provide more obvious message for instance:
"{relationshipName}[{index}].{keyName} is required"

src/Nox.Generator/Application/Commands/UpdateCommand.template.cs
```diff
@@ -107,4 +120,25 @@ private async Task UpdateLocalizedEntityAsync({{entity.Name}}Entity entity, {{en
...
+
+public class Update{{entity.Name}}Validator : AbstractValidator<Update{{entity.Name}}Command>
+{
+    public Update{{entity.Name}}Validator()
+    {
+		{{- for ownedRelationship in entity.OwnedRelationships }}
+			{{- if ownedRelationship.WithMultiEntity }}
+				{{- relationshipName = GetNavigationPropertyName entity ownedRelationship }}
+				{{- key = ownedRelationship.Related.Entity.Keys | array.first }}
+					{{- if IsNoxTypeCreatable key.Type }}
```
gjorgji-grgovski-authoritypartners: Guid is a special case and we should be able to provide items without providing a Guid value as a key, and they should be just created like new entries.
We can have tests for Guid keys for these cases.

gjorgji-grgovski-authoritypartners (changes requested):

src/Nox.Lib/Extensions/FluentApiExtensions.cs
anna-naumova25: I would say other methods in this class also don't correspond with FluentApi. So maybe we can rename the whole class? Let's align it with @rochar

src/Nox.Lib/Infrastructure/Persistence/EntityDbContextBase.cs
anna-naumova25: Agree, I've separated Repository and DbContext

src/Nox.Generator/Application/Commands/UpdateCommand.template.cs
anna-naumova25: Removed

src/Nox.Generator/Application/Factories/EntityFactory.template.cs
anna-naumova25: I've added CountryOwnedHoliday in ClientApi.Tests with Guid key to test all the different scenarios

src/Nox.Generator/Application/Commands/UpdateCommand.template.cs
anna-naumova25: Updated

src/Nox.Generator/Application/Commands/UpdateCommand.template.cs
anna-naumova25: Have update this and some other validations for Guid

gjorgji-grgovski-authoritypartners (approved):

gjorgji-grgovski-authoritypartners (approved):

src/Nox.Abstractions/Domain/IRepository.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    /// <summary>
+    /// Deletes Owned Entity
+    /// </summary>
+    void DeleteOwned<T>(T entity) where T : IOwnedEntity;
+
+    /// <summary>
+    /// Deletes a range of Owned Entities
+    /// </summary>
+    void DeleteOwnedRange<T>(IEnumerable<T> entities) where T : IOwnedEntity;
```
rochar, line 19: The name and the parameters, overload function, are sufficient 
```suggestion
    void DeleteOwned<T>(IEnumerable<T> entities) where T : IOwnedEntity;
```

src/Nox.Lib/Domain/Repository.cs
```diff
@@ -0,0 +1,30 @@
...
+    public void Delete<T>(T entity) where T : IEntity
+    {
+        _dbContext.Remove(entity);
+    }
+    public void DeleteOwned<T>(T entity) where T : IOwnedEntity
+    {
+        _dbContext.Remove(entity);
+    }
+    public void DeleteOwnedRange<T>(IEnumerable<T> entities) where T : IOwnedEntity
+    {
```
rochar, line 25: There is a Remove Range
https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.removerange?view=efcore-7.0

src/Nox.Lib/Domain/Repository.cs
rochar, line 1: The DbContext is the Repository, what is the advantage of having this level of abstraction?

src/Nox.Generator/Application/ServiceCollectionExtensions.template.cs
```diff
@@ -41,6 +41,12 @@ public static IServiceCollection AddNox(this IServiceCollection services, WebApp
             configurator.WithMessagingTransactionalOutbox<AppDbContext>();
             configureNox?.Invoke(configurator);
         });
+        services.AddScoped(typeof(Nox.Domain.IRepository), serviceProvider =>
```
rochar, line 44: this is enough, since all are registered in the container
```suggestion
        services.AddScoped<Nox.Domain.IRepository,Nox.Domain.Repository>
```

rochar (changes requested):
looks good @anna-naumova25 , just a few points to discuss

src/Nox.Lib/Domain/Repository.cs
gjorgji-grgovski-authoritypartners: @rochar This was actually my comment to split it, and there were a couple of reasons behind that like Separation of Concerns and the Single Responsibility Principle, as well maybe the main one that came to my mind is that like this we do not depend on the current ORM (EF Core), but we can also support tomorrow different ORM or maybe we would need to support some non-SQL database (even MongoDB can be used today with EF Core).
But if you think it should not be a separate layer of abstraction we can remove it.

src/Nox.Lib/Domain/Repository.cs
anna-naumova25: Right, I tried to use it, but for our "owned" entities it doesn't work as expected. So if I use
`_dbContext.RemoveRange(entities)` - it throws an exception like "the type of List< CountryLocalNames > is not registered in dbContext".

It should probably be used like `_dbContext.CountryLocalNames.RemoveRange(entitiesToRemove);`. But we don't have owned entities set available in dbContext, it only works from a parent.

src/Nox.Generator/Application/ServiceCollectionExtensions.template.cs
anna-naumova25: it is thrown the following exception if I change it to the suggestion:
`Unable to resolve service for type 'Microsoft.EntityFrameworkCore.DbContext' while attempting to activate 'Nox.Domain.Repository'`

src/Nox.Abstractions/Domain/IRepository.cs
anna-naumova25: Updated in https://github.com/NoxOrg/Nox.Generator/pull/1206

PR #1190, opened 2023-11-30
add IWritable for compound types

rochar:
register ComplexTypes before EntityType or Sets
change compound nox types dto to class


src/Nox.Generator/Application/Dto/NoxTypeDto.template.cs
```diff
@@ -7,13 +7,18 @@
 namespace {{codeGeneratorState.ApplicationNameSpace}}.Dto;
 
 {{~ for noxType in compoundTypes ~}}
+{{~ if noxType.NoxType != "LatLong2" ~}}
```
anna-naumova25: was it something for test purpose? I think we don't have LatLong2 nox type

src/Nox.Generator/Application/Dto/NoxTypeDto.template.cs
```diff
@@ -7,13 +7,18 @@
...
-public record {{noxType.NoxType}}Dto: I{{noxType.NoxType}}
+public class {{noxType.NoxType}}Dto: I{{noxType.NoxType}}, IWritable{{noxType.NoxType}}
 {    
+    #pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+    public {{noxType.NoxType}}Dto(){ } 
+    #pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+
     public {{noxType.NoxType}}Dto(            
             {{- for component in noxType.Components -}}
             {{component.UnderlyingType}} {{ToLowerFirstChar component.Name}}{{- if component.Name != lastName -}},{{- end -}}
```
anna-naumova25: Here you can use `{{if !for.last}},{{end}}` instead of `lastName`

anna-naumova25 (changes requested):
Looks good to me, just 2 small comments

src/Nox.Generator/Application/Dto/NoxTypeDto.template.cs
gjorgji-grgovski-authoritypartners: I guess this is just a leftover?

src/Nox.Types/Types/DateTimeRange/IDateTimeRange.cs
```diff
@@ -5,4 +5,9 @@ public interface IDateTimeRange
 {
     DateTimeOffset End { get; }
     DateTimeOffset Start { get; }
+}
+public interface IWritableDateTimeRange
```
gjorgji-grgovski-authoritypartners, line 9: I think to be clear we can also rename the existing Interfaces to be IReadable....
This can be the same for each compound type.

rochar: I will do it on a next PR if that is ok

gjorgji-grgovski-authoritypartners (approved):

PR #1192, opened 2023-11-30
Feature/294 add seed data

DanielHodgeGH:
Apologies about the size of this PR - I'll make sure the finishing Dataseeder set is smaller - please ignore any generated file code

This PR includes most of the DataSeeders for Cryptocash - including a special Unit test to convert raw Reference Currency JSON data into DataSeed Nox.Type validated Currency JSON

The Dataseeder fires if the table related to the Entity doesnt have any rows


samples/Cryptocash.Api/DataSeed/DataSeedExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            var dataSeeders = scope.ServiceProvider.GetServices<IDataSeeder>();
+
+            foreach (var dataSeeder in dataSeeders)
+            {
+                dataSeeder.Seed();
+            }
+        }
+        catch
+        {
+            // don't throw
```
rochar, line 48: I think it would be prefer to not catch nothing and let the exception to be raised, otherwise we can be mislead on errors that are related with corrupted data

anna-naumova25 (approved):

rochar (approved):
looks good @DanielHodgeGH 
can y please see my comment and have a new PR for it if necessary

PR #1193, opened 2023-11-30
NOX-257 Generate Enum Class for Enumeration Values

emir-sehovic1:


gjorgji-grgovski-authoritypartners (approved):

rochar (approved):
Looks good, I will merge

but consider to test NoxCodeConventions.GetEnumPropertyName since we can make sure that the rules are not broken and well applied

PR #1219, opened 2023-12-07
Make entity factor create update async

rochar:
add ensure referencenumber for reference number attributes


src/Nox.Lib/Infrastructure/Persistence/EntityDbContextBase.cs
```diff
@@ -54,6 +54,11 @@ public override async Task<int> SaveChangesAsync(CancellationToken cancellationT
 
         public virtual async Task<long> GetSequenceNextValueAsync(string sequenceName)
         {
+
+            var t = new List<int>();
```
gjorgji-grgovski-authoritypartners: I guess this is a leftover from some testing?

gjorgji-grgovski-authoritypartners (dismissed):

anna-naumova25 (dismissed):

src/Nox.Lib/Infrastructure/Persistence/EntityDbContextBase.cs
rochar: yes will remove it

PR #1220, opened 2023-12-07
feature 294 Initial commit for Country DataSeeder

DanielHodgeGH:
PR to add Country Dataseeder to Cryptocash.API sourced from Reference data json

Please note:

I have added a special Unit test which filters the raw reference Country json data and creates the JSON ready for the dataseeder - users dont need to run this test as I have already included the output Country json with this PR - so the unit test is set to unused currently

It tries a best attempt to convert all required Nox.Types from source data into DTO ready values - so please ignore the try catches as im deliberately doing that to use the Nox.Type internal validation when FROM is used so work out what are valid values - this makes sure the Country DTO list is in sync against the required foreign keys - ie: Currency for example

When dataseeding (on a new fresh database) it takes about 4mins to finish - currently adding about 220 Countries and about 8000 holidays.


DanielHodgeGH:
as discusssed - Added Fact Skip and usage comments on Country and Currency tests used when creating Dataseeder JSON files

rochar (approved):

anna-naumova25 (approved):

PR #1221, opened 2023-12-07
NOX-380 Update Get for Related Entity to return empty if no related is found and 404 if main entity not found

anna-naumova25:
Please note that for *OrOne relationships we use SingleResult response type and there is no easy way to return an empty SingleResult from those endpoints as its type is OData IQueryable.

https://github.com/NoxOrg/Nox.Generator/blob/79b5c99e805b72341fc3a509f3d91449b57ccfb6/src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData/WorkplacesController.g.cs#L113

So in the endpoint above the response will be either NotFound (when workplace is not found and when workplace.Country is null) or a CountryDto. 

Let's discuss if this needs to be changed.


rochar (dismissed):

src/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Presentation.Api.OData/WorkplacesController.g.cs
```diff
@@ -263,12 +265,13 @@ public virtual async Task<ActionResult> PostToTenants([FromRoute] System.Int64 k
     [EnableQuery]
     public virtual async Task<ActionResult<IQueryable<TenantDto>>> GetTenants(System.Int64 key)
```
rochar, line 268: we need to improve the performance, for example:

```
      var exists = await _mediator.Send(new GetWorkplaceByIdQuery(key))).Any();
      if(!exists)
      {
          return NotFound();
      }
```
and for single result we may not need to test if there are any results.

gjorgji-grgovski-authoritypartners (dismissed):

rochar (approved):

PR #1240, opened 2023-12-12
Add refernce number validations

rochar:
add EntityNotFoundException
remove package from dependantbot

### DIFF
--- a/src/Nox.Solution/Models/Domain/Domain.cs
+++ b/src/Nox.Solution/Models/Domain/Domain.cs
@@ -34,6 +34,8 @@ public class Domain : YamlConfigNode<NoxSolution, NoxSolution>
     {
         var result = base.Validate(topNode, parentNode, yamlPath);
 
+        ValidateUniqueReferenceNumberPrefixForAllEntitites(result);
+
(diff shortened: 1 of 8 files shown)


src/Nox.Solution/Models/Domain/Domain.cs
```diff
@@ -55,4 +57,32 @@ private void ValidateOnlyOneAutoNumberPerEntityForSqlLite(NoxSolution topNode, V
             }
         }
     }
+    private void ValidateUniqueReferenceNumberPrefixForAllEntitites(NoxSolution topNode, ValidationResult result)
```
gjorgji-grgovski-authoritypartners:
```suggestion
    private void ValidateUniqueReferenceNumberPrefixForAllEntitites(ValidationResult result)
```

gjorgji-grgovski-authoritypartners: I guess we do not need the NoxSolution here?

anna-naumova25 (dismissed):

gjorgji-grgovski-authoritypartners (approved):

PR #1243, opened 2023-12-12
NOX-409 Put to OwnedEntity should Create if not found

anna-naumova25:


src/Nox.Generator/Application/Commands/UpdateOwnedCommand.template.cs
```diff
@@ -104,7 +122,20 @@ internal partial class Update{{relationshipName}}For{{parent.Name}}CommandHandle
...
 		return new {{entity.Name}}KeyDto({{primaryKeysReturnQuery}});
-	}	
+	}
+	
+	private async Task<{{entity.Name}}Entity> CreateEntityAsync({{entity.Name}}UpsertDto upsertDto, {{parent.Name}}Entity parent)
+	{
+		var entity = await _entityFactory.CreateEntityAsync(upsertDto);
+		{{- for key in entity.Keys ~}}
+		{{- if key.Type == "Nuid" }}
+		entity.Ensure{{key.Name}}();
```
rochar: @anna-naumova25  we could do the Ensure Part for Nuid at the factory, for create and update (needs to be the last method to be call because depends on other properties), and remove this from all commands. We are already doing the same for reference number

rochar (dismissed):
looks good

just a suggestion to consider

src/Nox.Generator/Application/Commands/UpdateOwnedCommand.template.cs
anna-naumova25: @rochar I've double checked, we already have EnsureId int the owned entities factories. So I just removed those calls from the commands.

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):

PR #1246, opened 2023-12-13
Feature/nox 408 json source

jan-schutte:


src/Nox.Integration/Exceptions/NoxIntegrationTypeConversionException.cs
```diff
@@ -0,0 +1,13 @@
+namespace Nox.Integration.Exceptions;
+
+[Serializable]
```
rochar, line 3: deserialization constructor is mandatory for [Serializable]

rochar (changes requested):
looks good, one suggestion

src/Nox.Integration/Exceptions/NoxIntegrationTypeConversionException.cs
jan-schutte: Nice, thankx

PR #1247, opened 2023-12-13
NOX-416 Change Create and Update Dto make all Fields nullable

anna-naumova25:


src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
```diff
@@ -42,11 +48,7 @@ public abstract class {{className}}Base : IEntityDto<DomainNamespace.{{entity.Na
     /// <remarks>Required.</remarks>    
     [Required(ErrorMessage = "{{key.Name}} is required")]
     {{- end }}
-    {{ if key.Type == "EntityId" -}}
-    public {{SingleKeyPrimitiveTypeForEntity key.EntityIdTypeOptions.Entity}} {{key.Name}} { get; set; } = default!;
-    {{- else -}}
-    public {{SinglePrimitiveTypeForKey key}} {{key.Name}} { get; set; } = default!;
-    {{- end}}
+    public {{keyPrimitiveType key}}? {{key.Name}} { get; set; }
```
gjorgji-grgovski-authoritypartners:
```suggestion
    public virtual {{keyPrimitiveType key}}? {{key.Name}} { get; set; }
```

gjorgji-grgovski-authoritypartners (dismissed):
Looks good, just a small comment to make the key property virtual in the EntityCreateDto

rochar (dismissed):

src/Nox.Generator/Application/Dto/EntityCreateDto.template.cs
rochar: @anna-naumova25  I have applied the suggestion, so if builds I will merge it

PR #1252, opened 2023-12-14
NOX-414 Document in readme all our generated endpoints conventions and assumptions

anna-naumova25:


rochar (approved):

rochar:
@anna-naumova25  we should also add the the related explicitly /api/Entity/{key}/{RelatedEntity}.... but we can do this on the context of the issue for multiple levels of relations

PR #1260, opened 2023-12-18
NOX-415 Replace NotFound to EntityNotFoundException

anna-naumova25:
for RelatedEntities and OwnedEntities endpoint it will be done in scope of NOX-417


rochar (dismissed):

src/Nox.Generator/Presentation/Api/OData/EntityControllerGeneratorBase.cs
```diff
@@ -47,4 +47,10 @@ protected static string GetPrimaryKeysRoute(Entity entity, NoxSolution solution,
 
         return string.Empty;
     }
+
+    protected static string GetPrimaryKeysToString(Entity entity, string prefix = "key")
```
gjorgji-grgovski-authoritypartners: @anna-naumova25 I'm afraid I don't see the usage of this method.

anna-naumova25: right, thanks @gjorgji-grgovski-authoritypartners, I've removed it

PR #1261, opened 2023-12-18
NOX-417 improve command responses

anna-naumova25:


src/Nox.Lib/Exceptions/DatabaseSaveException.cs
```diff
@@ -0,0 +1,25 @@
...
+    public DatabaseSaveException(string message, Exception inner)
+        : base(message, inner)
+    {
+    }
+
+    public HttpStatusCode? StatusCode => HttpStatusCode.InternalServerError;
+
+    public string ErrorCode => "no_records_were_saved";
+
+    public object? ErrorDetails => throw new NotImplementedException();
```
rochar: we will serialize and add to the response this, so if no Details we should set it to null

src/Nox.Lib/Exceptions/DatabaseSaveException.cs
```diff
@@ -0,0 +1,25 @@
+using System.Net;
+
+namespace Nox.Exceptions;
+
+[Serializable]
```
rochar: We need to implement the deserialization constructor whenever we set the [Serializable] attribute

src/Nox.Lib/Exceptions/DatabaseSaveException.cs
```diff
@@ -0,0 +1,25 @@
...
+public class DatabaseSaveException : Exception, IApplicationException
```
rochar: may be call this DataNotSavedException? This does not represent a database exception per si...

src/Nox.Lib/Exceptions/DatabaseSaveException.cs
```diff
@@ -0,0 +1,25 @@
...
+        : base(message)
+    {
+    }
+
+    public DatabaseSaveException(string message, Exception inner)
+        : base(message, inner)
+    {
+    }
+
+    public HttpStatusCode? StatusCode => HttpStatusCode.InternalServerError;
```
rochar: I'm not sure if this can actually occur, the only reason I can remember is the etag is not valid... we need to discuss if the is a server error or not?

src/Nox.Lib/Exceptions/EntityAttributeIsNotNullableException.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Nox.Exceptions;
+
+/// <summary>
+/// trying to set a null to a non nullable property
+/// </summary>
+public class EntityAttributeIsNotNullableException : Exception
```
rochar: Can we delete this exception, I'm not seeing any usage. I know its not related with your PR, but lets improve this ? :)

rochar (changes requested):
looks good, some point for review

src/Nox.Lib/Exceptions/DatabaseSaveException.cs
rochar, line 1: All this exception are not specific with WebApi but Command Exceptions, so we can use this PR and move all them to Nox.Core.Exceptions?

anna-naumova25: done

src/Nox.Lib/Exceptions/EntityAttributeIsNotNullableException.cs
anna-naumova25: sure, it's removed now

anna-naumova25:
@rochar I've checked the case with passing wrong ETag and it fails on SaveChangesAsync in this case. So I don't know any use case of results<1. Therefore I removed that code and DatabaseSaveException. Let me know if you have any comments on this.

rochar (approved):

src/Nox.Generator/Application/Commands/DeleteByIdCommand.template.cs
```diff
@@ -58,7 +64,7 @@ public virtual async Task<bool> Handle(Delete{{entity.Name}}ByIdCommand request,
 			var entity = await DbContext.{{entity.PluralName}}.FindAsync({{entity.Keys | array.map "Name" | keysQuery}});
 			if (entity == null{{if (entity.Persistence?.IsAudited ?? true)}} || entity.IsDeleted == true{{end}})
```
gjorgji-grgovski-authoritypartners, line 65:
```suggestion
			if (entity == null)
```
This has nothing to do with your PR, but just to bring it here I don't think we need this, because there is a global filter in the AppDbContext that would never return soft deleted entries.

gjorgji-grgovski-authoritypartners (approved):
Approved with a small comment to discuss it with @rochar

PR #1262, opened 2023-12-18
NOX-388 owned entity localization factory

msadikkoseaupa:


src/Nox.Core/Domain/IEntityConcurrent.cs
```diff
@@ -3,7 +3,7 @@
 /// <summary>
 /// Entities whose updates need to be handled with optimistic concurrency
 /// </summary>
-public interface IEntityConcurrent
+public interface IEntityConcurrent: IEntity
```
rochar: We should avoid have inheritance on contracts.

src/Nox.Core/Domain/IRepository.cs
```diff
@@ -1,8 +1,17 @@
 ﻿
+using System.Linq.Expressions;
+
 namespace Nox.Domain;
 
 public interface IRepository
 {
+    // needs to work with IEntity, IOwnedEntity and localized entity
+    IQueryable<T> Query<T>(Expression<Func<T, bool>> predicate) where T : class, IEntity;
+    IQueryable<T> QueryOwned<T>(Expression<Func<T, bool>> predicate) where T : class, IOwnedEntity;
```
rochar: we  do not need the Owned, I guess it will not work since we do not have any set registered for owned

rochar (changes requested):

rochar (dismissed):

gjorgji-grgovski-authoritypartners (dismissed):

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #1267, opened 2023-12-19
add nox generator tasks executed as a test

anna-naumova25:
this is a very rough implementation, we can clean it up and use it if we're okay about the test itself


samples/Cryptocash.Api.Tests/CryptocashTasksTests.cs
```diff
@@ -0,0 +1,26 @@
+using Nox.Generator.Tasks;
+
+namespace Cryptocash.Tests;
+
+public class CryptocashTasksTests
+{
```
rochar, line 6: Add #if Debug, we can skip this on CI

rochar (approved):

PR #1272, opened 2023-12-20
NOX-432 initial PR for RelatedEndpointsMiddleware

anna-naumova25:


src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+            return;
+        }
+
+        if (!requestPath.StartsWithSegments(_apiPrefix))
+        {
+            await _next(context);
+            return;
+        }
+
+        if (context.Request.Method == "PATCH")
```
gjorgji-grgovski-authoritypartners:
```suggestion
        if (HttpMethods.IsPatch(context.Request.Method))
```

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            if (!ValidateSegments(segments))
+            {
+                await _next(context);
+                return;
+            }
+
+            var newPath = BuildNewPatchPath(segments);
+            context.Request.Path = newPath;
+            context.Request.QueryString = context.Request.QueryString;
```
gjorgji-grgovski-authoritypartners: Do we need this, as you set the same value?

src/Nox.Lib/Configuration/NoxUseOptions.cs
```diff
@@ -93,7 +93,8 @@ public void Configure(IApplicationBuilder builder)
         {
             appBuilder.UseMiddleware<VersionMiddleware>();
         });
-        builder.UseMiddleware<NoxApiMiddleware>().UseRouting();        
+        builder.UseMiddleware<RelatedEndpointsMiddleware>().UseRouting();
```
rochar: the use routing is redundant applies to the IApplication Builder, to be more explicitly you can do:

`builder
.UseMiddleware<RelatedEndpointsMiddleware>()
.UseMiddleware<RelatedEndpointsMiddleware>()
.UseRouting();`

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+
+    private readonly IEnumerable<string> _navigationNames;
+
+    public RelatedEndpointsMiddleware(RequestDelegate next, NoxSolution solution)
+    {
+        _next = next;
+
+        _apiPrefix = solution.Presentation.ApiConfiguration.ApiRoutePrefix;
+
+        if (solution.Domain is null)
```
rochar: makes sense, I would suggest then on the Use Registration do not even register the middleware, this can be applied to both middlewares

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+                .Distinct()
+                .ToList();
+        }
+    }
+
+    public async Task InvokeAsync(HttpContext context)
+    {
+        var requestPath = context.Request.Path;
+
+        if (!requestPath.HasValue)
```
rochar, line 34: to remove this to if/branch, I would suggest on the middleware registration to do something like:
`builder.UseWhen(context => context.Request.Path.StartsWithSegments("solution.Presentation.ApiConfiguration.ApiRoutePrefix")`

This can be applied to both middleware

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+        await _next(context);
+        return;
+    }
+
+
+    private List<string> ParsePath(string path)
+    {
+        List<string> segments = new List<string>();
+
+        if (path.StartsWith(_apiPrefix, StringComparison.OrdinalIgnoreCase))
```
rochar: this is redundant, should not get here if it does not start with api route prefix?

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        if (path.StartsWith("/", StringComparison.Ordinal)) 
+        {
+            path = path.Substring(1);
+        }
+
+        ReadOnlySpan<char> pathSpan = path.AsSpan();
+        int startIndex = 0;
+
+        while (startIndex < pathSpan.Length)
```
rochar, line 65: IMO we need to tradeoff everything and have a good performance

We should do the validate while parsing

For example:

1. for the fisrt segment you can test the _entitiesPluralNames...
2.  for each odd segment you can test the _navigationNames.Contains

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+
+internal class RelatedEndpointsMiddleware
+{
+    private readonly PathString _apiPrefix;
+
+    private readonly RequestDelegate _next;
+
+    private readonly IEnumerable<string> _entitiesPluralNames;
+
+    private readonly IEnumerable<string> _navigationNames;
```
rochar: Ienumerable hides the fact of doing a query every time/ or not... we should be explicit here

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+        _apiPrefix = solution.Presentation.ApiConfiguration.ApiRoutePrefix;
+
+        if (solution.Domain is null)
+        {
+            _entitiesPluralNames = Array.Empty<string>();
+            _navigationNames = Array.Empty<string>();
+        }
+        else
+        {
+            _entitiesPluralNames = solution.Domain.Entities.Select(e => e.PluralName).ToList();
```
rochar: we are using this for search consider using hashset instead

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+            _navigationNames = solution.Domain.Entities
```
rochar: we are using this for search consider using hashset instead

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+    private List<string> ParsePath(string path)
+    {
+        List<string> segments = new List<string>();
+
+        if (path.StartsWith(_apiPrefix, StringComparison.OrdinalIgnoreCase))
+        {
+            path = path.Substring(_apiPrefix.Value!.Length);
+        }
+
+        if (path.StartsWith("/", StringComparison.Ordinal))
```
rochar: it always start by /... do we need the comparation?

rochar (changes requested):

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: removed

src/Nox.Lib/Configuration/NoxUseOptions.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: updated

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):

PR #1285, opened 2023-12-22
Add ManageRelationshipDepth yaml property

anna-naumova25:


src/Nox.Solution/Models/Domain/Entities/Entity.cs
```diff
@@ -81,6 +81,11 @@ public class Entity : YamlConfigNode<NoxSolution, Domain>
     [AdditionalProperties(false)]
     public IReadOnlyList<UniqueAttributeConstraint> UniqueAttributeConstraints { get; internal set; } = Array.Empty<UniqueAttributeConstraint>();
 
+    [Title("Depth of relationship navigation.")]
+    [Description("Specifies the depth for generating related entities endpoints.")]
+    [Minimum(1), Maximum(5)]
+    public int ManageRelationshipDepth { get; internal set; } = 1;
```
rochar, line 87:
```suggestion
    public int ApiGenerateRelatedEndpointsMaxDepth { get; internal set; } = 1;
```

rochar: We need to think to move this to the API level with the CanManageEntity and CanManageRef

ApiGenerateRelatedEndpoint
ApiGenerateReferenceEndpoint

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #1286, opened 2023-12-22
add Money to Nox.Ui.Blazor.Lib

DanielHodgeGH:
add Money to Nox.Ui.Blazor.Lib

Note: added temp MoneyDTO / MoneyModel conversion methods to ListEntityPageBase so that that Add Form works for now - need to review when we come to integrate generator output

Also have used Format & Symbol info to handle currency display for now until we have something better using Currency Ref Data integrated into Nox.Ui.Blazor.Lib

Changed Nox.Ui.Blazor.Lib to ref Nox.Core instead of Nox.Types


rochar (dismissed):

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor.cs
```diff
@@ -0,0 +1,98 @@
+﻿using Microsoft.AspNetCore.Components;
+using Nox.Types;
+using Nox.Ui.Blazor.Lib.Models;
+
+namespace Nox.Ui.Blazor.Lib.Components.NoxTypes;
+
+public partial class EditMoney : ComponentBase
+{
+#nullable enable
```
anna-naumova25: This one can be removed as we have nullable enabled in csproj

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor
```diff
@@ -1,5 +1,5 @@
-﻿@using Cryptocash.Ui.Generated.Component;
-@inherits UiEditMoneyBase
+﻿@using MudBlazor
+@using static EditMoney
```
anna-naumova25: I think 'using static' only brings benefits when EditMoney class has static members. In our case it seems to be redundant for me.

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/ViewMoney.razor
```diff
@@ -0,0 +1,3 @@
+﻿@using static ViewMoney
```
anna-naumova25: same here, I think we don't need it as ViewMonay doesn't have any static members

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/ViewMoney.razor.cs
```diff
@@ -0,0 +1,58 @@
+﻿using Microsoft.AspNetCore.Components;
+using Nox.Ui.Blazor.Lib.Models;
+
+namespace Nox.Ui.Blazor.Lib.Components.NoxTypes;
+
+public partial class ViewMoney : ComponentBase
+{
+#nullable enable
```
anna-naumova25: already included in csproj

anna-naumova25 (changes requested):
Looks good, just some small comments

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor.cs
DanielHodgeGH: Removed in latest commit

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/ViewMoney.razor.cs
DanielHodgeGH: Removed in latest commit

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/ViewMoney.razor
DanielHodgeGH: similar to EditMoney

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor
DanielHodgeGH: Removed static and have change to Inherits latest commit - been forced to use UiEditMoney for class name instead as it clashes with razor class - cant seem to find a way around it - or would we prefer EditMoneyBase ? the class name I think needs to be different to the razor name

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor
```diff
@@ -1,5 +1,5 @@
-﻿@using Cryptocash.Ui.Generated.Component;
-@inherits UiEditMoneyBase
+﻿@using MudBlazor
+@inherits UiEditMoney
```
anna-naumova25: I think we don't need this line at all, as per naming convention the EditMoney will be automatically available in this file. However the class should be partial and should have the same name as the razor part.

So for this case we should have:
EditMoney.razor
EditMoney.razor.cs

In the first file we should not have `@inherits ...`, the model will be available automatically.
But there should be defined the namespace same as in EditMoney.razor.cs:
`namespace Nox.Ui.Blazor.Lib.Components.NoxTypes`

In the second class it should be `public partial class EditMoney: ComponentBase`

@DanielHodgeGH  please test this approach and update it for all the components if it works

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor.cs
```diff
@@ -0,0 +1,97 @@
+﻿using Microsoft.AspNetCore.Components;
+using Nox.Types;
+using Nox.Ui.Blazor.Lib.Models;
+
+namespace Nox.Ui.Blazor.Lib.Components.NoxTypes;
+
+public class UiEditMoney : ComponentBase
```
anna-naumova25:
```suggestion
public partial class EditMoney : ComponentBase
```

anna-naumova25 (changes requested):

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor
DanielHodgeGH: thanks Anna - tested okay have added to latest commit

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMoney.razor.cs
DanielHodgeGH: added on latest commit as discussed

anna-naumova25 (approved):

rochar (approved):
@DanielHodgeGH we already have the money and country extensions available see https://github.com/NoxOrg/Nox.Generator/commit/f8b7c644ae664fb3e6e243d8026fc6d7bca95df9

DanielHodgeGH:
> @DanielHodgeGH we already have the money and country extensions available see [f8b7c64](https://github.com/NoxOrg/Nox.Generator/commit/f8b7c644ae664fb3e6e243d8026fc6d7bca95df9)

thanks have created https://regusit.atlassian.net/browse/NOX-549 so I dont forget

PR #1287, opened 2023-12-22
NOX-526 rename related endpoints yaml properties

anna-naumova25:


src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -22,6 +22,20 @@ public RelatedEndpointsMiddleware(RequestDelegate next, NoxSolution solution)
...
+            /Address/Key/City/Key/Monuments/key
+        the dictionary for Cities will be the following
+            Capital, Cities
+            MajorCities, Cities
+            Cities, Cities
+            City, City
+        */
         foreach (var entity in solution.Domain!.Entities)
         {
             foreach(var relationship in entity.Relationships)
```
rochar, line 41: @anna-naumova25  we can also ignore Relationships that ApiCan* are both false

rochar (approved):

PR #1291, opened 2023-12-26
NOX-429 localization improvements

msadikkoseaupa:


src/Nox.Yaml/Enums/CultureCode/Culture.cs
gjorgji-grgovski-authoritypartners, line 1: @rochar "Create in Nox.Types.Abstraction an enumeration with all valid values (Namespace …/Enums/CultureCode)" - this is not satisfied, because we have circular dependencies if it has to be under Nox.Types.Abstraction, so I instructed @msadikkose to put it for now under Nox.Yaml and we can discuss this further and maybe even move the Enums under a separate project or review the current References.

samples/Cryptocash.Api.Tests/CryptocashTasksTests.cs
gjorgji-grgovski-authoritypartners, line 1: @rochar  let's put this as a Note for everyone that the Paths cannot be Windows-specific only, in this case, this is executed only in Debug mode, but we noticed the same thing on the Nox.Reference, so it is good to be documented.

@msadikkose I guess this change is because you are not using Windows locally and had to fix this code?

msadikkoseaupa: Yes, exactly!

src/Nox.Yaml/Schema/Generator/SchemaProperty.cs
```diff
@@ -414,37 +414,79 @@ private static string ToJsonSchemaType(Type? type)
     /// <returns>A list of JSON schema enumerators.</returns>
     private static List<string>? ToEnumValues(Type? type)
     {
-        if (type is null || !type.IsEnum) return null;
-
+        if (type is null ) return null;
+        Type? enumType = null;
+        if (!type.IsEnum && !IsEnumerableOfEnum(type, out enumType)) return null;
```
rochar, line 419: @msadikkoseaupa what is the reason to add support to IEnumerable of Enum? What is the use case?

src/Nox.Yaml/Enums/CultureCode/Culture.cs
rochar: whenever some one is creating a yaml definition/schema and would like to use our Nox.Yaml component (for example our Nox.Solution),  how could he do it without changing the Nox.Yaml? how can it be extended? Open Close principle?

Nox.Yaml can only know about Yaml, and not concrete knowledge about other domains.

CultureCode is a Nox.Type concept, so should live in Nox.Types.Abstraction, and is used by Nox.Solution.

src/Nox.Types/Types/CultureCode/CultureCode.cs
```diff
@@ -7,6 +14,13 @@ namespace Nox.Types;
 /// </summary>
 public partial class CultureCode : ValueObject<string, CultureCode>
 {
+    private static readonly ImmutableHashSet<string> SupportedCultureCodes = System.Enum.GetValues(typeof(Culture)).Cast<Culture>().Select(v =>
```
rochar: this is being computed in other place, we could make this an extension method to CultureCode... eventually keep it in a static field, so we have only a single instance in memory

src/Nox.Types/Types/CultureCode/CultureCode.cs
```diff
@@ -15,14 +29,41 @@ internal override ValidationResult Validate()
...
+    {
+       
+        var field =  value.GetType().GetField(value.ToString());
+        DisplayNameAttribute? attribute = null;
+        if(field is not null)
+        {
+            attribute = field.GetCustomAttributes(typeof(DisplayNameAttribute), false).FirstOrDefault() as DisplayNameAttribute;
+        }
+        var cultureCode = attribute?.DisplayName ?? value.ToString();
+        var newObject = new CultureCode
```
rochar: statement from here to the end of the method  can be replaced by:
```suggestion
       return From(cultureCode);
```

src/Nox.Types/Types/CultureCode/CultureCode.cs
```diff
@@ -15,14 +29,41 @@ internal override ValidationResult Validate()
...
+    /// <summary>
+    /// Creates a new instance of <see cref="CultureCode"/>
+    /// <param name="value">The Culture value type of <see cref="Culture"/> to create the <see cref="CultureCode"/> with</param>
+    /// </summary>
+    /// <returns>A new instance of <see cref="CultureCode"/></returns>
+    public static CultureCode From(Culture value)
+    {
+       
+        var field =  value.GetType().GetField(value.ToString());
+        DisplayNameAttribute? attribute = null;
```
rochar: suggest to add the computation of the DisplayName/Name to an extension method and keep this metho more simple to read and maintain

Actually I already see that you have it, so use it :)

rochar (changes requested):

src/Nox.Yaml/Enums/CultureCode/Culture.cs
rochar: @gjorgji-grgovski-authoritypartners @msadikkoseaupa we can discuss some solution, if you have any proposal just let me know it

samples/Cryptocash.Api.Tests/CryptocashTasksTests.cs
rochar: yes, we can add to our code review guidelines

src/Nox.Types/Types/CultureCode/CultureCode.cs
msadikkoseaupa: Desired changens have been applied

src/Nox.Types/Types/CultureCode/CultureCode.cs
msadikkoseaupa: Desired changens have been applied

src/Nox.Types/Types/CultureCode/CultureCode.cs
msadikkoseaupa: Desired changens have been applied

src/Nox.Yaml/Enums/CultureCode/Culture.cs
msadikkoseaupa: Well if we have Nox.Enums project seperated we can create calculated constant dictionary, extensions there and it would be cleaner without any circular dependency. 
What do you think @gjorgji-grgovski-authoritypartners @rochar ?

src/Nox.Types/Types/CultureCode/CultureCode.cs
rochar: @msadikkoseaupa I'm not seeing the change?

`public static CultureCode From(Culture value)
{
   return From(value.ToDisplayName());
}`

src/Nox.Types.Abstractions/Types/CultureCode.cs
```diff
@@ -1,7 +1,22 @@
...
-    public class CultureCode
+    public static class CultureCode
     {
         public const string RegularExpression = @"^[a-z]{2}(?:-[A-Z]{2})?(?:-[A-Z][a-z]{3})?$";
+        
+        public static readonly ImmutableSortedDictionary<string,Culture> CultureCodeDisplayNames = Enum.GetValues(typeof(Culture))
+            .Cast<Culture>()
+            .ToImmutableSortedDictionary(x =>
+                {
+                    var field = x.GetType().GetField(x.ToString())!;
```
rochar: Use the extension method...

```suggestion
                    x.ToDisplayName();
```

rochar (changes requested):
looks good, just a couple of suggestions

src/Nox.Types.Abstractions/Types/CultureCode.cs
msadikkoseaupa: Desired changens have been applied

src/Nox.Types/Types/CultureCode/CultureCode.cs
msadikkoseaupa: Sorry for the missunderstanding. Desired changens have been applied

rochar (dismissed):

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #1292, opened 2023-12-26
NOX-526 Add validation by ManageRelationshipDepth, CanManage*

anna-naumova25:


src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -45,6 +50,8 @@ public RelatedEndpointsMiddleware(RequestDelegate next, NoxSolution solution)
                 {
                     _navigationNameToEntityPluralName[navigationName] = relationship.EntityPlural;
                 }
+
+                _canRedirect.Add((entity.PluralName.ToLower(), navigationName.ToLower()), relationship.ApiGenerateRelatedEndpoint);
```
rochar: we could ignore the if the relationship.ApiGenerateRelatedEndpoint is false:

 foreach(var relationship in entity.Relationships)
 {
     if(!elationship.ApiGenerateRelatedEndpoint) continue;

so is not part of the dictionary.... so the dictionary only contains the valid ones...

rochar (dismissed):

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: Updated as per suggestion

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):

PR #1294, opened 2023-12-27
NOX-527 apply redirection changes for all verbs for related endpoints

anna-naumova25:
Will clean up PATCH related code in the next PR.


src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
```diff
@@ -166,4 +176,19 @@ private PathString BuildNewPatchPath(List<string> segments)
...
     }
+
+    private PathString BuildNewPath(List<string> segments)
+    {
+        int count = segments.Count;
+        bool isEvenCount = count % 2 != 0;
+
+        if (isEvenCount)
+        {
+            return new PathString(_apiPrefix + "/" + _navigationNameToEntityPluralName[segments[count - 3]] + "/" + segments[count - 2] + "/" + segments[^1]);
```
rochar: @anna-naumova25  I would suggest to have the _apiPrefix + "/" precomputed in some variable (or add the / to the _apiPrefix when its read from the configuration), avoiding creating an extra string every time...

rochar (dismissed):

src/Nox.Lib/Middlewares/RelatedEndpointsMiddleware.cs
anna-naumova25: have updated _apiPrefix

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #1305, opened 2023-12-29
PR to migrate Temperature, MacAddress, Percentage and Number

DanielHodgeGH:


src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMacAddress.razor.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        return new PatternMask(Format)
+        {
+            MaskChars = new[] { new MaskChar('#', @"[0-9a-fA-F]") },
+            CleanDelimiters = true,
+            Transformation = AllUpperCase
+        };
+    }
+
+    private static char AllUpperCase(char c) => c.ToString().ToUpperInvariant()[0];
```
rochar: char defines a to upper method ` Char.ToUpper(ch, culture)` and avoid create a string for this...

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/ViewTemperature.razor.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            return String.Empty;
+        }
+    }
+
+    private string DisplayTemperatureUnit()
+    {
+        if (TemperatureUnit == TemperatureUnit.Fahrenheit)
+        {
+            return " " + TemperatureUnit.Fahrenheit.Symbol;
```
rochar: I would suggest this "Empty Space" prefix to be added on the DisplayTemperature fproperty, since this one should return only the units... this can be difficult to maintain in the future

rochar (commented):
looks good, two suggestions to review

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/EditMacAddress.razor.cs
DanielHodgeGH: amended in latest commit

src/Nox.Ui.Blazor.Lib/Components/NoxTypes/ViewTemperature.razor.cs
DanielHodgeGH: amended in latest commit - plus Area, Length, Distance as well

rochar (approved):

anna-naumova25 (approved):

PR #1331, opened 2024-01-08
Add generator task to Cryptocash.Ui

anna-naumova25:


anna-naumova25:
@rochar please note that build of the Solution fails with this approach until we kill MSBuild. So probably everyone who works with Nox solution will face with this issue (not only people who work on Ui).

DanielHodgeGH (approved):

rochar (approved):

PR #1347, opened 2024-01-12
Improve SQL query for relationship chain validator

anna-naumova25:
@rochar  please see the generated SQL query in NOX-214 ticket's comments.


rochar (approved):
good job

PR #1348, opened 2024-01-12
NOX-290 Delete Entity and Owned Localized Translations

emir-sehovic1:


src/Nox.Lib/Extensions/IQueryableExtensions.cs
rochar, line 1: All code that is not WebApi related should be moved to Nox.Core

rochar (changes requested):

src/Nox.Lib/Extensions/IQueryableExtensions.cs
rochar: actually I'm not finding any usage of this extension?

rochar (changes requested):

src/Nox.Lib/Extensions/IQueryableExtensions.cs
emir-sehovic1: @rochar It is used in the template [src/Nox.Generator/Application/Commands/DeleteOwnedEntityTranslationsCommand.template.cs](https://github.com/NoxOrg/Nox/pull/1348/files#diff-e48fdae6336ecd13e726b2301880db8c10475e2c5d40f5c72320244eee87428c:~:text=%7B%7Bentity.PluralName%7D%7DLocalized.WhereAnyMatch)
at line 74

rochar: ok thks, it's on the Scriban Template, that why I did not find it!

src/Nox.Core/Exceptions/EntityNotFoundException.cs
```diff
@@ -12,6 +12,12 @@ public EntityNotFoundException(string entityName, string entityId)
         ErrorDetails = new { Entity = entityName, Key = entityId };
     }
 
+    public EntityNotFoundException(string entityName, string entityId, string cultureCode)
```
rochar: In this case we did not found the entity or the localized text for that culture
IMO we should create a specific exception for this, so its clear that we did not found the localized text, for example: `EntityLocalizedTextNotFound`

src/Nox.Generator/Application/Commands/DeleteOwnedEntityTranslationsCommand.template.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        await OnCompletedAsync(command, entity);
+
+        DbContext.Remove(entity);
+        {{~else~}}
+        await DbContext.Entry(parentEntity).Collection(p => p.{{GetNavigationPropertyName parent relationship}}).LoadAsync(cancellationToken);
+        {{if entity.Keys.size > 1}}
+        var entityKeys = parentEntity.{{GetNavigationPropertyName parent relationship}}.Select(x => new { {{for key in entity.Keys}} {{key.Name}} = x.{{key.Name}}{{if !for.last}}, {{end}}{{end}} }).ToList();
+        var entities = await DbContext.{{entity.PluralName}}Localized.WhereAnyMatch(entityKeys, (x,y) => {{for key in entity.Keys}}x.{{key.Name}} == {{key.Name}} && {{end}}x.CultureCode == command.CultureCode).ToListAsync(cancellationToken);
```
rochar: can you add an example to ClientApi.tests, so we have a owner entity related to Many Owned Entity, that contains localized text, and is using WhereAnyMatch? thks

tests/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands/DeleteTenantBrandsTranslationsForTenantCommand.g.cs
```diff
@@ -0,0 +1,83 @@
...
+    public virtual async Task<bool> Handle(DeleteTenantBrandsTranslationsForTenantCommand command, CancellationToken cancellationToken)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        await OnExecutingAsync(command);
+		var parentKeyId = ClientApi.Domain.TenantMetadata.CreateId(command.keyId);
+        var parentEntity = await DbContext.Tenants.FindAsync(parentKeyId);
+
+        if (parentEntity is null)
+        {
+            throw new EntityNotFoundException("Tenant", $"{parentKeyId.ToString()}");
```
rochar: I would suggest to add a Guard to EntityNotFoundException to do this, and avoid the branching on the code... I know we are doing this in other parts, but this a is a good opportunity to improve.

EntityNotFoundException.ThrowIfNull(parentEntity, "Tenant", $"{parentKeyId.ToString()}")

src/Nox.Generator/Application/Commands/DeleteOwnedEntityTranslationsCommand.template.cs
rochar: We do not support this use case, for example, changing TenantBrands key to be a composite key:

`//  - The Nox type validation failed with 1 error(s). PropertyName: TenantBrands. Error: Relationship [TenantBrands] on entity [Tenant] refers to related entity [TenantBrand] with composite key. Must be simple key on TenantBrand.   at Nox.Yaml.YamlConfigurationReader`2.ValidateAll(TFullType instance) in ....YamlConfigurationReader.cs:line 239`

rochar (changes requested):

src/Nox.Generator/Application/Commands/DeleteOwnedEntityTranslationsCommand.template.cs
emir-sehovic1: @rochar  yes, currently compound keys are not supported when dealing with owned relationships, however I noticed that rest of the templates were written without that assumption, I suppose for purpose of easier upgrade if this feature will be supported at some point. So I opted to write this template in that way as well. Let me know if you would still prefer me to delete this case.

rochar: I would prefer to follow the Yagni principle in this case. Because we do not have any plans to support this.

anna-naumova25 (approved):

rochar (approved):

PR #1379, opened 2024-01-22
NOX-646 Fix Swagger error for ApiRouteMappings

emir-sehovic1:


src/Nox.Solution/Models/Presentation/ApiConfiguration/ApiRouteMappings/ApiRouteMapping.cs
```diff
@@ -74,13 +78,38 @@ public override ValidationResult Validate(NoxSolution topNode, ApiConfiguration
...
+            if (!definedParameters.Any(x => string.Equals(x, parameter, StringComparison.InvariantCultureIgnoreCase)))
+            {
+                result.AddError(nameof(RequestInput), $"Endpoint [{Name}] defines a parameter [{parameter}] in the route that is not defined in the [{nameof(RequestInput)}] section.");
+            }
+        }
+
         return result;
     }
 
+    private IEnumerable<string> ExtractParametersFromRoute()
```
rochar: I would suggest to return an IReadonlyCollection, so its clear that the result is not going to be computed
```suggestion
    private IReadOnlyCollection<string> ExtractParametersFromRoute()
```

src/Nox.Solution/Models/Presentation/ApiConfiguration/ApiRouteMappings/ApiRouteMapping.cs
```diff
@@ -74,13 +78,38 @@ public override ValidationResult Validate(NoxSolution topNode, ApiConfiguration
...
+    {
+        string parametersPattern = @"\{(\w+)}";
+        MatchCollection matches = Regex.Matches(Route, parametersPattern, RegexOptions.None, TimeSpan.FromMilliseconds(100));
+
+        List<string> parameterNames = new();
```
rochar: matches.Count is the expected size of the list?
```suggestion
        List<string> parameterNames = new(matches.Count);
```

src/Nox.Lib/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Nox.Extensions;
+
+internal static class StringExtensions
+{
+    public static string StripQueryParameters(this string input) => input.Split('?')[0];
```
rochar: this approach will require less memory allocation?
```suggestion
    public static string StripQueryParameters(this string input) =>input.Substring(0, input.IndexOf(',?'));
```

src/Nox.Lib/Extensions/StringExtensions.cs
rochar, line 1: Although this seems a string extension, is actually more a URI extension, I would suggest to move this method to ApiRouteMappingDocumentFilter.cs and make it private for now

rochar (changes requested):
looks good, just a couple suggestions

emir-sehovic1:
> looks good, just a couple suggestions

@rochar Thank you for the feedback, I've applied the changes

anna-naumova25 (approved):

rochar (approved):

PR #1389, opened 2024-01-25
add Selection List component

DanielHodgeGH:


src/Nox.Ui.Blazor.Lib/Components/Generic/SelectList.razor.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+
+    #region Declarations
+
+    [Parameter]
+    public List<SelectEntityModel> EntityList { get; set; } = new();
+
+    public string? CurrentIdList { get; set; }
+
+    public string? DisplaySelectedList { get; set; }
```
anna-naumova25, line 18:
```suggestion
    public string? DisplaySelectedList => EntityList.Any() ? string.Join(", ", EntityList.Select(Entity => Entity.Name)) : string.Empty;
```

src/Nox.Ui.Blazor.Lib/Components/Generic/SelectList.razor.cs
```diff
@@ -0,0 +1,79 @@
+﻿using Microsoft.AspNetCore.Components;
+using Microsoft.VisualBasic;
+using MudBlazor.Charts;
+using Nox.Ui.Blazor.Lib.Models;
+
+namespace Nox.Ui.Blazor.Lib.Components.Generic;
+
+public partial class SelectList : ComponentBase
```
anna-naumova25: Maybe I'd rename it to SelectMany. @rochar what do you think?

anna-naumova25 (dismissed):
Looks good, just one small comments that can be done in the next PR.

src/Nox.Ui.Blazor.Lib/Components/Generic/SelectList.razor
```diff
@@ -0,0 +1,26 @@
...
+               FullWidth="true"
+               Clearable="true"
+               Required="@Required"
+               RequiredError="@ErrorRequiredMessage(Title)"
+               TextChanged="OnIdChanged"
+               Value="@CurrentIdList"
+               Text="@DisplaySelectedList">
+        @foreach (SelectEntityModel item in SelectionList)
+        {
+            if (!string.IsNullOrWhiteSpace(item.Id)
```
rochar: @DanielHodgeGH  this safe guard can hide or is hidding some issue in the domain

can not exist an entity without an Id or a text field, we should not support such cases

I would prefer to remove the safe guard and see if we need to fix anything in other place?

src/Nox.Ui.Blazor.Lib/Components/Generic/SelectList.razor.cs
rochar: maybe MultiSelect?

SelectMany is more about flatten a lists of lists

src/Nox.Ui.Blazor.Lib/Components/Generic/SelectList.razor.cs
```diff
@@ -0,0 +1,79 @@
...
+        CurrentIdList = newValue;
+        IEnumerable<string> selectedIdList = new List<string>();
+        if (!string.IsNullOrWhiteSpace(CurrentIdList))
+        {
+            selectedIdList = CurrentIdList.Split(',').Select(Id => Id.Trim()).ToList();
+        }
+
+        EntityList = new();
+
+        if (selectedIdList.Any())
```
rochar: this is unnecessary, the foreach does this implicitly?

src/Nox.Ui.Blazor.Lib/Components/Generic/SelectList.razor.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            foreach (string currentId in selectedIdList)
+            {
+                SelectEntityModel? entity = SelectionList.Find(Entity => 
+                !string.IsNullOrWhiteSpace(Entity.Id)
```
rochar: the same here, actually at the moment we are not adding an entity if the id is empty...

rochar (changes requested):
looks good, a few comments to review

anna-naumova25 (approved):

rochar (approved):

PR #1396, opened 2024-01-29
NOX-692 - Add Nox Type Extensions for CurrencyCode / CurrencyCode3

dejobratic:
1. Update Nox.Type.CurrencyCode3 to use CurrencyCodeEnum  (Maybe build the hashset on the static constructor, using the enum?)
2. Create CurrencyCode Extension and move MoneyExtensions. GetReferenceCurrency(this CurrencyCode currencyCode) to it
   a. Add a new extension public static CurrencyCode GetCurrencyCode(this Currency referenceCurrency)
   b. Add a new extension public static CurrencyCode3 GetCurrencyCode3(this Currency referenceCurrency)
3. Consolidate Codes In both sides (CurrencyCode Enum and Reference Currency). If there is a conflict use Nox.Reference has the source of truth


tests/Nox.Types.Extensions.Tests/Types/CurrencyCode/CurrencyCodeTests.cs
```diff
@@ -0,0 +1,134 @@
...
+        referenceCurrency.Should().NotBeNull();
+        referenceCurrency.Id.Should().Be(currencyId);
+        referenceCurrency.Name.Should().Be(currencyName);
+        referenceCurrency.IsoCode.Should().Be(currencyIsoCode);
+        referenceCurrency.IsoNumber.Should().Be(currencyIsoNumber);
+        referenceCurrency.Symbol.Should().Be(currencySymbol);
+    }
+
+    [Fact(Skip = "There are 29 missing Nox.Reference.Currency")]
+    public void WhenGettingReferenceCurrency_FromAllCurrencyCodes_Success()
```
dejobratic, line 74: Following 29 currency codes cannot be found in Nox.Reference:
SSP
SLE
VED
UYW
VES
MRU
STN
ZWL
UYI
CHE
CHW
XBA
XBB
XBC
XBD
XAU
XDR
XAG
XPT
XTS
XPD
XUA
COU
MXV
BOV
CLF
XSU
USN
XXX

Since Sonar is failing when we skip the test, should we remove it or fix the Nox.Reference data before merging this one?

rochar: @dejobratic you can suppress that warning in the code for now. So we can integrate this one.

And meanwhile y should update Nox.Reference and then un skip this tests.

samples/Cryptocash.Ui/Nox.Generated/Generator.g.cs
rochar, line 1: Manual build Nox.Generator.Task... and get rid of this error.

src/Nox.Types/Types/CurrencyCode3/CurrencyCode3.cs
```diff
@@ -57,7 +38,7 @@ internal override ValidationResult Validate()
     {
         var result = base.Validate();
 
-        if (!_currencyCodes.Contains(Value))
+        if (!Enum.TryParse<CurrencyCode>(Value, out _))
```
rochar: for performance reasons  in the static constructor build an hashset of valid string using the CurrencyCode Enum 

we should measure the difference, between Enum.TryParse and the hashset contains

rochar (changes requested):
looks good, a couple of suggestions

src/Nox.Types/Types/CurrencyCode3/CurrencyCode3.cs
gjorgji-grgovski-authoritypartners: Suggestion: We can have an EnumExtension for the conversion of Enum values into HashSet<string> because we should be able to reuse it in another place

rochar (dismissed):

src/Nox.Types/Types/CurrencyCode3/CurrencyCode3.cs
dejobratic: Yes, HashSet should be faster because of constant access time, since we don't have many entries.
Changes have been applied accordingly.

src/Nox.Types/Common/EnumExtensions.cs
rochar, line 1: @dejobratic 
look to this bench mark
image

not sure if can be applied to the items in Enum.GetValue...

rochar (dismissed):

src/Nox.Types/Common/EnumExtensions.cs
dejobratic: Thanks for the suggestion. Let me check the benchmark and I'll let you know my findings.

dejobratic: Results:
image

Setup:
image

So, looks like using nameof is faster, even though we are creating a static hash set, so I'm changing the implementation.

gjorgji-grgovski-authoritypartners (dismissed):

src/Nox.Types/Common/EnumExtensions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Nox.Common;
+
+public static class EnumExtensions
+{
+    public static HashSet<string> ToHashSet<T>() where T : struct, Enum
+        => Enum.GetValues<T>().Select(e => nameof(e)).ToHashSet();
```
dejobratic: I've rushed a bit into conclusion. We cannot use nameof(e) when dynamically creating a hash set.. 

We can either dynamically create the hash set using .ToString() or explicitly provide the full list of nameof(CurrencyCode.*). The first approach does not require any code change when we add a new currency code, whereas the second one does.

These are the benchmark results after the change
image


Which approach should we take, @rochar?

rochar: any of the 2 bests would be ok

rochar: we can use the first one, not a big deal, its just done once

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

tests/Nox.Types.Extensions.Tests/Types/CurrencyCode/CurrencyCodeTests.cs
gjorgji-grgovski-authoritypartners: This would be covered with https://regusit.atlassian.net/browse/NOX-752
CC: @rochar @dejobratic

PR #1399, opened 2024-01-29
NOX-657 add entity services

anna-naumova25:


DanielHodgeGH (dismissed):
Thanks @anna-naumova25 looks good - the only thing for me - which we need to discuss is the GetAllAsync - I think should use SEARCHFILTER and ORDERBY options to build on the fly GET ODATA querys to filter and order the returning dataset - but this needs to be agreed for a future PR probably - thanks DanHo

src/Nox.Generator.Tasks/Ui/Extensions/ServiceCollectionExtension.scriban-cs
```diff
@@ -0,0 +1,22 @@
...
+using Nox.Ui.Blazor.Lib.Services;
+using {{codeGenConventions.UiNameSpace}}.Services;
+
+namespace {{codeGenConventions.UiNameSpace}}.Extensions;
+
+public static class {{className}}
+{
+    public static IServiceCollection AddNoxUi(this IServiceCollection services)
+    {
+        services.AddSingleton<GlobalDataService>();
```
rochar: @anna-naumova25 lets rename GlobalDataService, I always need to open it to understand what is the goal of it :)

maybe?
```suggestion
        services.AddSingleton<ApplicationState>();
```

src/Nox.Generator.Tasks/Ui/Extensions/ServiceCollectionExtension.scriban-cs
```diff
@@ -0,0 +1,22 @@
...
+        services.AddMudServices();
```
rochar, line 11: @anna-naumova25  can you take a look at this https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

and review if we can make this more resilient?

samples/Cryptocash.Ui/Nox.Generated/Ui.Extensions/ServiceCollectionExtension.g.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Cryptocash.Ui.Extensions;
+
+public static class ServiceCollectionExtension
+{
+    public static IServiceCollection AddNoxUi(this IServiceCollection services)
+    {
+        services.AddSingleton<GlobalDataService>();
+        services.AddMudServices();
+        services.AddHttpClient();
+        services.AddScoped<EndpointsProvider>();
```
rochar: this one can be singleton? never changes?

samples/Cryptocash.Ui/Nox.Generated/Ui.Services/EndpointsProvider.g.cs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public string BaseUrl
+    {
+        get { return _baseUrl; }
+    }
+
+    public string BaseUrlWithRoutePrefix
+    {
+        get { return $"{BaseUrl}/api"; }
```
rochar, line 41: We should consider not creating a new string every time, in this case IMO it's preferable to keep the value in Memory and return always the same instance. What do y thin?

rochar (changes requested):

samples/Cryptocash.Ui/Nox.Generated/Ui.Services/EndpointsProvider.g.cs
```diff
@@ -0,0 +1,76 @@
+﻿// Generated
+
+namespace Cryptocash.Ui.Services;
+
+public class EndpointsProvider
```
rochar: Should Abstract, and request the Interface in the DI

anna-naumova25:
> Thanks @anna-naumova25 looks good - the only thing for me - which we need to discuss is the GetAllAsync - I think should use SEARCHFILTER and ORDERBY options to build on the fly GET ODATA querys to filter and order the returning dataset - but this needs to be agreed for a future PR probably - thanks DanHo

Let's see if we can use Simple.OData.Client
https://www.odata.org/blog/advanced-odata-tutorial-with-simple-odata-client/
https://www.nuget.org/packages/Simple.OData.Client/

samples/Cryptocash.Ui/Nox.Generated/Ui.Services/EndpointsProvider.g.cs
anna-naumova25: as discussed, will be done in scope of NOX-650

src/Nox.Generator.Tasks/Ui/Extensions/ServiceCollectionExtension.scriban-cs
anna-naumova25: updated

src/Nox.Generator.Tasks/Ui/Extensions/ServiceCollectionExtension.scriban-cs
anna-naumova25: updated

samples/Cryptocash.Ui/Nox.Generated/Ui.Extensions/ServiceCollectionExtension.g.cs
anna-naumova25: updated

samples/Cryptocash.Ui/Nox.Generated/Ui.Services/EndpointsProvider.g.cs
anna-naumova25: updated

rochar (approved):

DanielHodgeGH (approved):

PR #1407, opened 2024-01-31
NOX-662 consolidate postalcodes

msadikkoseaupa:


src/Nox.Types/Types/StreetAddress/CountryPostalCodeValidator.cs
```diff
@@ -53,198 +54,159 @@ public static bool IsValid(string countryCode, string? postalCode)
...
-        ("UZ",@"^\d{3} \d{3}$"),
-        ("VA",@"^120$"),
-        ("VE",@"^\d{4}(\s[a-zA-Z]{1})?$"),
-        ("VN",@"^\d{6}$"),
-        ("WF",@"^986\d{2}$"),
-        ("ZM",@"^\d{5}$")
-    });
+    private static readonly ImmutableList<(string CountryCode, string Pattern)> PostalCodesMapping =
+        ImmutableList.CreateRange(new[]
+        {
```
rochar, line 60: @msadikkoseaupa  Besides consolidation, we should keep the ones that do not exist in Nox.Reference and add them to Nox.Reference as well, correct?

rochar (changes requested):
@msadikkoseaupa just a comment for align expectations

src/Nox.Types/Types/StreetAddress/CountryPostalCodeValidator.cs
msadikkoseaupa: Actually I did not merge them, instead I completely replaced them from Reference. 
Let me update it. 
So what if same country code has multiple regex from multiple resources

rochar (approved):

gjorgji-grgovski-authoritypartners (approved):

PR #1412, opened 2024-02-01
NOX-698 Update Nox.Solution Messaging Server making possible use Az Manage Identity

emir-sehovic1:


tests/Nox.Solution.Tests/YamlFileValidationTests.cs
```diff
@@ -42,6 +42,44 @@ public void When_AzureServiceBus_AzureServiceBusConfig_Is_Required()
         Assert.Throws<NoxYamlValidationException>(() => noxConfigBuilder.Build());
     }
 
+    [Fact]
```
dejobratic, line 45: Can we also add an additional test where both SharedAccessKey and SharedAccessKeyName are set?

dejobratic (commented):
Looks good. Just a comment related to tests.

dejobratic (approved):

rochar (approved):

PR #1413, opened 2024-02-01
Example for using TypeOptions in NoxTypes Ui Components

anna-naumova25:
Changes in AddCountryForm.razor will be overwritten by Task run. It is just for an example.


rochar (commented):
yes looks good we can follow this approach

rochar (approved):

DanielHodgeGH (approved):

PR #1421, opened 2024-02-06
NOX-761 Adjust UI components for NoxType.CurrencyNumber, NoxType.CountryCode3, NoxType.CountryNumber

anna-naumova25:


DanielHodgeGH (commented):
I understand we need Number and Codes but this is missing the point of the UI in my opinion as we need a bespoke instead of  global list of Currency and Country list for the Cryptocash example - I think we need to discuss further

rochar:
> I understand we need Number and Codes but this is missing the point of the UI in my opinion as we need a bespoke instead of global list of Currency and Country list for the Cryptocash example - I think we need to discuss further

One thing is editing an attribute of an entity, that the type is CurrencyCode or CountryNumber...

Another thing is modeling a Country/Currency Entity tailored for the Current Domain (Different data, different attributes, different rules). Here we are managing the Country Entity and not the attribute.

IMO I think you are mixing two different concepts one is the Entity other is the Attribute Type. We do not design for specific entities.

rochar (dismissed):

DanielHodgeGH (dismissed):
ok agreed so for Country and Country in Cryptocash we use Select component with ID and Name - cc: @rochar

PR #1468, opened 2024-03-11
NOX-849 - Update Put Entity behavior for OneOrMany Owned Entities

dejobratic:
1. Fixed issue when deleting all owned entities where relationship type is oneOrMany -> deletion should not be possible.
2. Renamed methods for handling owned entitites (Create, Update, Delete) -> removed RefTo from name.


rochar (approved):

src/Nox.Generator/Application/Commands/UpdateOwnedManyCommand.template.cs
```diff
@@ -71,6 +71,9 @@ internal partial class {{className}}HandlerBase : CommandCollectionBase<{{classN
 		var parentEntity = await _repository.FindAndIncludeAsync<{{codeGenConventions.DomainNameSpace}}.{{parent.Name}}>(keys.ToArray(),e => e.{{relationshipName}}, cancellationToken);
 		EntityNotFoundException.ThrowIfNull(parentEntity, "{{parent.Name}}",  "{{keysToString parent.Keys}}");
 
+		{{- if relationship.Relationship == "ZeroOrMany" }}
```
gjorgji-grgovski-authoritypartners, line 74: Why do we need this if block?

gjorgji-grgovski-authoritypartners (approved):

PR #1475, opened 2024-03-15
NOX-820 - Delete Owned Entity with Localized field fails in Postgres

dejobratic:


tests/Nox.ClientApi.Tests/Generated/Nox.Generator/Nox.Generator.NoxCodeGenerator/Application.Commands/DeleteCountryLocalNamesForCountryCommand.g.cs
```diff
@@ -46,7 +46,7 @@ public virtual async Task<bool> Handle(DeleteCountryLocalNamesForCountryCommand
 		
 		var keys = new List<object?>(1);
 		keys.Add(Dto.CountryMetadata.CreateId(request.ParentKeyDto.keyId));
-		var parentEntity = await Repository.FindAndIncludeAsync<ClientApi.Domain.Country>(keys.ToArray(), p => p.CountryLocalNames, cancellationToken);
+		var parentEntity = await Repository.FindAndIncludeAsync<ClientApi.Domain.Country, ClientApi.Domain.CountryLocalName, ClientApi.Domain.CountryLocalNameLocalized>(keys.ToArray(), p => p.CountryLocalNames, p => p.LocalizedCountryLocalNames, cancellationToken);
```
rochar: This is what we want to avoid, localized is not part of the "Domain", it should cascade delete on the DB

we need to fix the root cause, related to duplicate field/foreign key in localized table

dejobratic: The duplicate foreign key was introduced because we configured the localized entity relationship as unidirectional, which was the case when they were introduced. But in the meantime, the relationship was changed to bidirectional, by adding the localized entities property on the entity. So now the fix was added by registering the localized entity relationship as bidirectional.

dejobratic: image

dejobratic: image

gjorgji-grgovski-authoritypartners (approved):

rochar (approved):
nice job

PR #1486, opened 2024-05-14
make Actions Ui Component generic defaulting to Add control

DanielHodgeGH:
the main file change is src/Nox.Ui.Blazor.Lib/Components/Generic/Actions.razor - the rest is just the Nox.Ui rewriting the files from the template - I have done it so that to the control defaults to an ADD control but you can change that by Title and Icon optionally - thanks


jan-schutte (approved):
looks good

PR #1522, opened 2024-07-31
MP-11521 - [ENHANCE-582] [Nox] Vuln Remediation

dejobratic:
- Resolved vulnerabilities on top level and transitive packages for all projects in solution.
- Due to issue with test run on `ClientApi.Tests` where tests were running for a long time (possibly indefinitely, because it took a few hours after which the run was manually canceled) xunit package(s) were downgraded for a minor and patch version, which solved the issue.
- Tests inside `Nox.Ui.Blazor.Lib.Tests.EditTextTests` are marked as skipped, because they were throwing ``System.InvalidOperationException : Cannot provide a value for property 'Localizer' on type 'MudBlazor.MudInput`1[[System.String, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]'. There is no registered service of type 'MudBlazor.InternalMudLocalizer``


DanielHodgeGH:
not sure about this review - as I dont like mudblazor being rolled back to a previous version plus over 1000 files changes - @gjorgji-grgovski-authoritypartners can you please review first thanks

dejobratic:
> not sure about this review - as I dont like mudblazor being rolled back to a previous version plus over 1000 files changes - @gjorgji-grgovski-authoritypartners can you please review first thanks

Thanks for the comment.

I've consolidated Mudblazor version to the latest one (maybe I've rolled it back by mistake).
For the 1000+ files - 99.9% of them are generated code, which for some reason were removed in some of the previous PRs.

DanielHodgeGH (approved):

